PR #2, opened 2016-02-16
Cjh/inital move over

sudocrystal:
Sets up files for transferring in documents on cs-fundamentals and leadership and inclusion.


sudocrystal:
@jnf @theutahkate Just letting you know I'm merging this in. Basically no content here. Just structure setup.

PR #14, opened 2016-08-16
Update 02a-data-structures.md

susanev:
Typo on line 93.
Also, was confused by the Let's Try It Out! section, I assumed it was supposed to be 1,2,3 instead of 1,1,1 .. but maybe I'm missing something.


sudocrystal:
Markdown numbering of ordered lists allows you to type 1. 1. 1. instead of actually numbering items so that it auto numbers. Though sometimes the auto number gets wonky with paragraphs in there.

PR #16, opened 2016-08-30
Fixed each example's result

jmojennifer:
Going over Enumerables just now, I noticed that the [1 ,2 ,3].each loop iterating to do n ** 2 was showing [1, 4, 9] as the results in the markdown file. I remembered Kari saying in class each won't accomplish this and map/collect is necessary. (Verified in IRB that the code results in [1, 2, 3] before making pull request; wanted to save someone some time by making the pull request myself.)


kariabancroft:
Since the `each` loop is changing the `results` Array inside the loop - the results array will be updated with these values. The following code is what would _not_ produce the desired results:
`results = [1,2,3].each { |n| n ** 2 }`

PR #18, opened 2016-09-09
Html css build (Day One)

PilgrimMemoirs:
Day one Lecture Material.


05-html-css/01-the-web.md
```diff
@@ -0,0 +1,74 @@
+# 🌐 Intro to the web 🌐
+
+
+## 📚 Learning Goals 📚
```
sudocrystal: Is this an appropriate place to also talk about how the Internet actually works? I like this video: https://www.youtube.com/watch?v=PBWhzz_Gn10 and then an activity that I showed Kari a couple cohorts ago where you use notecards/pieces of paper to send messages around the room and simulate lost packets, packet #s, etc.  ....too late to add this in? @PilgrimMemoirs @kariabancroft @CheezItMan

05-html-css/04-content-hierarchy.md
```diff
@@ -0,0 +1,118 @@
...
+    </p>
+  </article>
+  <footer>
+    <h3> &copy; 2016 </h3>
+  </footer>
+</body>
+```
+
+
+### Draw a relationship tree for the html code below:
```
sudocrystal: I like this activity!

PR #37, opened 2016-09-20
Moving Content from Sinatra to Rails

CheezItMan:
@kariabancroft 

I'm currently working on 06-ERB-and-Layouts.md, but making good progress.  Take a look if it's ok.  I'll finish the rest after my next driving lesson, pray for me ;).  

I'll do another pass through it all in the early morning as well.


08-rails/05-routes-controllers-and-views.md
```diff
@@ -42,3 +47,64 @@ Let's break down this code:
...
+<h1>Books#index</h1>
+<p>Find me in app/views/books/index.html.erb</p>
+
+</body>
+</html>
+```
+
+### The Controller Class
+
+Ruby places controller classes in the `app/controllers` folder.  In the case of the Controller we just generated it will be in the `books_controller.rb` class.
```
kariabancroft: Rails, not Ruby

08-rails/05-routes-controllers-and-views.md
```diff
@@ -42,3 +47,64 @@ Let's break down this code:
...
+Rails can go ahead and create a Controller for us with the following commands.
+
+```bash
+bin/rails generate controller Books index
+```
+
+This command has Rails generate a Controller and Views for the /books path, so when you bring up http://localhost:3000/index you will get the following HTML file.  
+
+```html
+<!DOCTYPE html>
```
kariabancroft: I think this file info might just be more confusing in here - maybe just describe what the page looks like?

08-rails/06-ERB-and-Layouts.md
```diff
@@ -50,37 +50,40 @@ Both code tags and value tags can be combined to create loops with Ruby code tha
...
 
 ```ruby
-# in our Sinatra app
-get "/products" do
-  @all_products = Product.all #returns an array of Product instances
-  erb :products # this tells Sinatra to render views/products.erb
+# in our controller app/controllers/books_controller.rb
+def index
+  @books = ["Once and Future King", "Fellowship of the ring",
+              "Ruby: How to Program", "Web Design with HTML, CSS, JavaScript"]
```
kariabancroft: It might be nice to have something here about the implicit render of the view with the corresponding name (or include an explicit render here)

PR #39, opened 2016-09-20
adds a new rails lecture demonstrating the user of OmniAuth/OAuth

jnf:
This mergers three discussions in daily curriculum. It deprecates teaching `has_secure_password` in rails in favor of teaching OAuth as the primary means of authentication. I think it should be followed with a livecode demonstrating the addition of login/logout links to views.


kariabancroft:
Looks good - thank you for consolidating!

PR #51, opened 2016-10-04
Just Added a quick way to reset the db between seeds

CheezItMan:
Useful for when seeds go wrong.

Also added a section on resource routing.


kariabancroft:
I think it would be great to add exactly the steps that db:reset executes, :drop, :create, :migrate, :seed

PR #52, opened 2016-10-05
Move Daily Curriculum Resources

kariabancroft:
- Create an intermediate rails folder
- Copy/pasted a ton of daily curriculum resources into this new folder
- Added todos for some resources
- Added learning goals for some resources

TODO: I still want to determine if we need/want to talk about all of these topics, but bringing them in will help us be on the same page to have those conversations


CheezItMan:
@kariabancroft It looks good to me.

PR #59, opened 2016-10-11
Controller Testing

kariabancroft:
- Controller testing
- A bit of grammar updates


CheezItMan:
@kariabancroft Just a question, I was under the impression that we would do spec-style testing from our conversation.  I'm happy to revert to assert-style in my model notes, just need to go back through things.

PR #68, opened 2016-10-12
note about rails foundation gem, to not use

PilgrimMemoirs:
added note about foundation rails gem


kariabancroft:
Looks good! Thanks for adding this

PR #84, opened 2016-10-19
Add information about *when* to use View Helpers

droberts-ada:
Including a nice table!

I know we just talked about having stuff ready *before* the night before, but I got a little excited while I was doing prep.

Edits to code are just indentation so that the Atom markdown preview would do the right thing.


kariabancroft:
Looks good!

PR #87, opened 2016-10-20
Small changes to interview prep whiteboarding questions

Hamled:
Most of these commits are just organizational changes, but the following are my attempt to make the questions more clear and provide a more specific progression of difficulty for the students:

* 719e071 - WB questions: Reorder for increasing difficulty
* 49148cc - WP questions: Improve clarity on lists questions

Those should probably receive some scrutiny because my assumptions about what questions are more difficult, and especially my interpretation of what the questions are asking, could be wrong.

### DIFF
--- a/04-cs-fundamentals/interview-prep/interview-questions-algorithms.md
+++ b/04-cs-fundamentals/interview-prep/interview-questions-algorithms.md
@@ -9,9 +9,9 @@
 1. Reverse words in a string (words are separated by one or more spaces).  e.g. "This is stuff" --> "stuff is This" _Extra fun: Do it in-place._
 
 ### Arrays
-1. You are given an array with integers between 1 and 1,000,000. One integer is in the array twice. Determine which one.
-1. You are given an array with integers between 1 and 1,000,000. One integer is missing. How can you determine which one? _Extra fun: Can you think of a way to do it while iterating through the array only once._
-1. Return the sum of the two largest integers in an array e.g. Given [19, 4, 2, -19, 100, 9], return 119 (119)
+1. You are given an unsorted array with integers between 1 and 1,000,000. One integer is in the array twice. Determine which one. _Extra fun: Do it without sorting._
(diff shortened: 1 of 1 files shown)


04-cs-fundamentals/interview-prep/interview-questions-algorithms.md
```diff
@@ -1,15 +1,27 @@
...
+1. Reverse a string e.g. `"Hello"` --> `"olleH"`
+1. Reverse words in a string (words are separated by one or more spaces).  e.g. `"This is stuff"` --> `"stuff is This"` _Extra fun: Do it in-place._
+
+### Arrays
+1. You are given an unsorted array with integers between 1 and 1,000,000. One integer is in the array twice. Determine which one. _Extra fun: Do it without sorting._
+1. You are given an unsorted array containing all integers between 1 and 1,000,000. One integer is missing. How can you determine which one? _Extra fun: Can you think of a way to do it by iterating through the array only once, without sorting?_
+1. Return the sum of the two largest integers in an array e.g. Given `[19, 4, 2, -19, 100, 9]`, return `119` (100 + 19)
+1. Return the largest sum of contiguous integers in the array e.g. Given `[-10, 2, 3, -2, 0, 5, -15]`, the largest sum is `8` (2 + 3 + -2 + 0 + 5)
 
 <!-- UNSURE IF ANY OF THESE ARE ANY GOOD TO SHARE WITH STUDENTS - will look at later.
```
Hamled: Apparently I cannot comment directly on lines that are "outside" the changes for this PR, but consider these to be comments about the questions below this line (use the button on the left side of the bottom line of the file view to see further context).

#### Prison guard:
* This seems like a pretty advanced graph question. If we didn't have the symmetric requirement for friendship, then it might be a much more tractable tree problem.
* With that requirement it still might be good, but we'd probably have to clarify if cyclical friendships are possible. Maybe have them try to solve in both cases.
* The "problem domain" or context for this given problem is, I think, problematic given Ada's social justice mission.

#### Tape Backup System:
* This seems pretty good and I think we could add it in at an intermediate level.

PR #96, opened 2016-10-28
Add textbook section on consuming an api in rails

droberts-ada:
We'll need to adjust code samples, etc. to match whatever live code starter package we provide.


CheezItMan:
That looks really good.

PR #99, opened 2016-11-01
Rework example tests

droberts-ada:
They now include the possibility of failure! Previously we just called
listchannel. Now we call sendmsg with both a positive and a negative
case.

In order to get this working with POST params, I had to make VCR
sensitive to the body of requests, in addition to the method and URI.

Also fixed a couple of grammar and formatting things.

### DIFF
--- a/09-intermediate-rails/08-vcr-testing.md
+++ b/09-intermediate-rails/08-vcr-testing.md
@@ -7,6 +7,7 @@
 
 ### How is testing APIs different?
 We know why testing is important, we have seen the importance of the TDD cycle. When testing APIs it is important to note that we are now relying on external sources for our own code to work. There are a few things to consider when testing external sources:
+
 1. If they're down, we're down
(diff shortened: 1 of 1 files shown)


09-intermediate-rails/08-vcr-testing.md
```diff
@@ -51,31 +53,32 @@ Each interaction is recorded as a **cassette**. We can load cassettes in our tes
...
   ```ruby
-  test "Retrieve the Channel data" do
+  test "Can send valid message to real channel" do
     VCR.use_cassette("channels") do
-      response = ApiWrapper.list_channels
-      assert response.length > 0
-      # Anything else contextual about API data?
+      message = "test message"
+      response = SlackApiWrapper.sendmsg("test-api-brackets", message)
+      assert response["ok"]
```
Hamled: Would it be worth it to have
```ruby
assert_not response["error"]
assert_respond_to response["message"], :[]
```
in this example?

PR #102, opened 2016-11-07
Move all information about Rails API optimizations down to an appendix

droberts-ada:


droberts-ada:
Whoops, looks like there's a little more on this. Don't merge yet.

PR #103, opened 2016-11-07
Last minute changes to the Producing APIs textbook page

droberts-ada:
- Move all information about API optimizations down to an appendix.
- Add section on `as_json` filters, since that's included in the lecture's TDD (Kari actually mentioned this in her notes, but we never got around to it).


kariabancroft:
Are we good to merge this?

PR #104, opened 2016-11-10
JavaScript stuff

kariabancroft:
- Updated some of the existing resources with typos and other random things
- Added resource for hoisting - not sure in what order we want to talk about this topic
- Added two new resources: split out DOM manipulation and eventing into two separate jQuery resources. We can do the first one (DOM manip) on Friday and the Eventing on Monday. Updated with a few new CodePen samples, so please take a look at those: 
- http://codepen.io/kariabancroft/pen/gLaBve
- http://codepen.io/kariabancroft/pen/woKYRR


CheezItMan:
Looks like a good start, more to do on jQuery but a good start.

10-JavaScript/03-objects-and-inheritance.md
```diff
@@ -1,12 +1,55 @@
...
+end
+
+FuzzyPet.NAP_HOURS  # 7
+FuzzyPet.dog        # 'borf!'
+```
+
+Now, let's try the same thing with JavaScript! hmmmm, how do we do this?
+
+```javascript
+var FuzzyPet = function(){
```
droberts-ada: Does this code work? I think FuzzyPet just wants to be an object, not a function.

10-JavaScript/08-jquery-events.md
```diff
@@ -0,0 +1,58 @@
...
+
+```javascript
+$("body").keydown(function(event) {
+  if (event.key == 'a') {
+    alert('You got an A!');
+  }
+});
+```
+
+## jQuery Documentation
```
droberts-ada: This section also exists in the other jQuery document

10-JavaScript/02-fun-with-functions.md
```diff
@@ -68,30 +68,58 @@ var backwards = function(x, y){
...
-  var args = Array.prototype.slice.call(arguments, 1);
-  console.log( "Hello " + name + ", " + args.join(' ') );
-};
+#### A note about `this`
+The keyword `this` is probably the most misunderstood concept in JavaScript. At its core, `this`, when invoked inside a function, refers to the invocation _context_ of the containing function (wat). Essentially, `this` is kinda like `self` in Ruby (this object, right here), but it is much, much more common in JavaScript for functions to be executed in contexts beyond where they were declared.
 
-hello( "Jeremy", "what", "are", "you", "doing" ); //=> "Hello Jeremy, what are you doing"
+As we explore how functions work, we will also discover the ins and outs of `this`. For now, we can think of `this` to mean _this object, right here_.
+
+```javascript
```
droberts-ada: This example feels sort of arbitrary, and sort of confusing. It definitely took me a couple of read-throughs to figure out what was going on.

It would be cool if we could come up with a more realistic / useful example, but I have no idea what that might be. Probably fine for now, so long as we're willing to spend some time on it.

PR #114, opened 2016-11-15
Textbook Updates: JS in the browser, jQuery 101

droberts-ada:
I intentionally did not use CodePen for this lecture, since it's more of a setup / "how do I do this in the real world" thing.


kariabancroft:
Do we need to re-order some of the other lectures to accommodate this one being #7?

CheezItMan:
Yes we do, moving Closures at least.

kariabancroft:
Sorry - I used the review feature and had no idea what I was doing - so my changes only now showed up

PR #115, opened 2016-11-16
Backbonejs

CheezItMan:
*very* raw backbone material.


kariabancroft:
Should we merge this in knowing it is a WIP so we can start collaborating on it?

Hamled:
Sounds like a good plan to me.

PR #119, opened 2016-11-17
Minor fix in setup instructions.

CheezItMan:


Hamled:
Can we have them just `cd` into the directory first?

This is a pretty inconsequential difference but it seems like the students are more familiar with the strategy of moving into a directory and then modifying its contents, rather than using commands with directory paths.

droberts-ada:
Yes.

PR #121, opened 2016-11-18
Ajax lessons

PilgrimMemoirs:
@kariabancroft


PilgrimMemoirs:
Still in progress, btw - though has overview. I'm going to run through now that I have the pets API hosted on Heroku and fill it in as I run though.

PR #122, opened 2016-11-19
Substantially expand the jQuery Events lecture

droberts-ada:
- Add motivation
- Expand discussion of callbacks and anonymous functions
- Add connection to `$(document).ready()`
- Talk about the `event` parameter
- Add in-class exercise via CodePen

### DIFF
--- a/10-JavaScript/07-jquery-events.md
+++ b/10-JavaScript/07-jquery-events.md
@@ -1,28 +1,43 @@
-# jQuery Eventing
+# jQuery Event Handling
+
+## Motivation
+- Users do things all the time: clicking on buttons, highlighting text, pressing keys
+- Being able to respond to these events is the core of making a web page dynamic
 
 ## Learning Goals
 - See jQuery events
 - Use jQuery events to manipulate the DOM
 
-## First, let's see an example
+## jQuery Event Handling
+### First, an Example
 Here is [an updated version](http://codepen.io/kariabancroft/pen/gLaBve) of the CodePen example we used with intro to jQuery.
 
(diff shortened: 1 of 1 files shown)


droberts-ada:
This lecture might still need more expansion. Two potential topics:
- Event delegation
- Multiple handlers for the same event

But, I figured I should get this up for review sooner rather than later.

10-JavaScript/07-jquery-events.md
```diff
@@ -1,28 +1,43 @@
...
+```javascript
+$(document).ready(function() {
+  // Do some stuff, manipulate the DOM
 });
 ```
 
-The jQuery `click()` function, waits for a user event--in this case the
-user clicking the element with their mouse--and then calls the
-function that you provide (the callback) when that event happens.
+We use this to make sure the HTML document has finished loading before we start messing with it. Turns out `ready()` is just another event. The only difference is it's triggered by the browser automatically rather than by something the user does.
```
kariabancroft: Nice correlation

10-JavaScript/07-jquery-events.md
```diff
@@ -41,3 +56,22 @@ $("body").keydown(function(event) {
   }
 });
 ```
+
+Notice the `event` argument to our callback. This will get filled in with a bunch of info about what happened to trigger the event - for example, which key was pressed, exactly when the event occurred, and where the mouse was on the screen when it happened. The `event` argument has actually been available for all our events so far, we've just been ignoring it.
+
+#### Exercise: Color Switching
+[Using this CodePen](http://codepen.io/droberts-ada/pen/pNRbJd), build a script that changes the background color of the color-box div each time the spacebar or enter key is pressed.
```
kariabancroft: Nice - I think this will be a good example

PR #123, opened 2016-11-20
Add two more sections to the jQuery Events lecture

droberts-ada:
- Section on binding multiple events and the use of `$(this)`
- Section on event delegations


droberts-ada:
I'm a little worried this will be a *TON* of content - maybe too much for one morning. Definitely something they should know about though.

Anyway, that's why I've got this as a separate PR.

10-JavaScript/07-jquery-events.md
```diff
@@ -68,10 +69,78 @@ Hints:
 - Detecting keys like space and enter is a little different than letter keys. Look it up!
 - `console.log()` is your friend.
 
+## Advanced Event Handling
+### Binding to Multiple Elements
```
kariabancroft: I think we do want this for the keyboard exercise - since each key shouldn't have a separate event handler

10-JavaScript/07-jquery-events.md
```diff
@@ -68,10 +69,78 @@ Hints:
...
+
+```javascript
+$(document).ready(function() {
+  $('li').click(function(event) {
+    alert('Got a click on an <li> containing "' + $(this).html() + '"');
+  });
+});
+```
+
+### Event Delegation
```
kariabancroft: This conversation we might be able to move out since it won't be as relevant for the keyboard assignment (though overall important and they will definitely encounter it at some point)

CheezItMan: Agreed.

PR #126, opened 2016-11-21
Hoisting

kariabancroft:
- Just a few basic changes
- Also update the var/const lecture even though I'm not sure that we will want to use it


kariabancroft:
I think by merging this I will mess up the overall numbering, but I can fix it

PR #129, opened 2016-11-22
Backbone Updates

kariabancroft:
- Just minor updates and clarifications to backbone notes
- Added a preliminary discussion resource for frameworks vs libraries (which could use some additions, maybe a video?)

### DIFF
--- a/11-Backbonejs/00-frameworks-vs-libs.md
+++ b/11-Backbonejs/00-frameworks-vs-libs.md
@@ -1,0 +1,25 @@
+# JavaScript: Frameworks vs. Libraries
+
+So far we have used one JavaScript library, jQuery! Next up we are going to use another JavaScript library. Oftentimes during our conversations about JavaScript we talk about using a framework or a library. There are a lot of opinions about which is better, and oftentimes even the opinion about if there is a meaningful difference between the two.
+
+Was Rails a Framework or a library? Framework! Let's see how our understanding of Rails falls under our new/refreshed understanding of Frameworks vs Libraries.
+
+## Frameworks
+A framework is a collection of patterns and libraries to help with building an application. A framework can decide upon and create structure.
+
+Examples:
+- React
+- Angular
+
+## Libraries
+A library is a collection of functions and/or objects that serve one particular purpose. You can simplify things that you create, but generally we do not use libraries to influence overall structure.
+
+Examples:
+- jQuery
+- Backbone
+
+## Why One Over The Other?
+To determine whether you want to use a library or a framework, you really need to this about the level of control you'd like to have over the application. 
+
+## Additional Resources
+- [Top JavaScript Frameworks, Libraries and Tools and When to Use Them](https://www.sitepoint.com/top-javascript-frameworks-libraries-tools-use/)
--- a/11-Backbonejs/00-frameworks-vs-libs.md
+++ b/11-Backbonejs/00-frameworks-vs-libs.md
@@ -17,9 +17,10 @@ A library is a collection of functions and/or objects that serve one particular 
 Examples:
 - jQuery
 - Backbone
+- Underscore
 
 ## Why One Over The Other?
-To determine whether you want to use a library or a framework, you really need to this about the level of control you'd like to have over the application. 
+To determine whether you want to use a library or a framework, you really need to this about the level of control you'd like to have over the application.
 
 ## Additional Resources
 - [Top JavaScript Frameworks, Libraries and Tools and When to Use Them](https://www.sitepoint.com/top-javascript-frameworks-libraries-tools-use/)
--- a/11-Backbonejs/01-Introduction-to-Backbonejs.md
+++ b/11-Backbonejs/01-Introduction-to-Backbonejs.md
@@ -1,67 +1,60 @@
-
-# Intro to Backbonejs
+# Intro to Backbone.js
 
 ## Learning Goals
 
 By the end of this lesson you should be able to:
 
-- Explain what Backbonejs provides for you as a developer
+- Explain what Backbone provides for you as a developer
 - Explain the MV* structure of Backbone
 
-## What is Backbonejs
+## What is Backbone
 
-Pure Javascript & JQuery can quickly become unstructured and messy.  It's very easy to build a single-page-application using Javascript & JQuery as a collection of Function calls, nested JQuery callbacks all tied to DOM elements in your HTML.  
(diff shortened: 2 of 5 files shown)


11-Backbonejs/00-frameworks-vs-libs.md
```diff
@@ -0,0 +1,25 @@
...
+Examples:
+- React
+- Angular
+
+## Libraries
+A library is a collection of functions and/or objects that serve one particular purpose. You can simplify things that you create, but generally we do not use libraries to influence overall structure.
+
+Examples:
+- jQuery
+- Backbone
```
Hamled: Lets add Underscore as well.

11-Backbonejs/01-Introduction-to-Backbonejs.md
```diff
@@ -1,67 +1,60 @@
...
+1.  **Collections**:  These are ordered sets of models
+1.  **Views**:  These render the models and listen for DOM/Model events.  Views can be used to render a single model to render or a collection.  
+1.  **Routers**:  Used for Single Page Applications, routers let you add paths to the URL so that you can maintain history and allow users to jump to a specific state in your single-page-application.  A router always contains at least one route and a function to map that particular route to.  
 
-![Backbone JS Arcitechture from https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&ved=0ahUKEwiknZLWnrDQAhULrVQKHdKdDtsQjxwIAw&url=http%3A%2F%2Fwww.slideshare.net%2Fronreiter%2Fwriting-html5-web-apps-using-backbonejs-and-gae&psig=AFQjCNHvjo7n-Xt_eytdroPc3zdRUVQY7g&ust=1479487463101423](images/backbonejs-architecture.jpg)
+![Backbone JS Architecture from https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&ved=0ahUKEwiknZLWnrDQAhULrVQKHdKdDtsQjxwIAw&url=http%3A%2F%2Fwww.slideshare.net%2Fronreiter%2Fwriting-html5-web-apps-using-backbonejs-and-gae&psig=AFQjCNHvjo7n-Xt_eytdroPc3zdRUVQY7g&ust=1479487463101423](images/backbonejs-architecture.jpg)
```
Hamled: Lets change this to say "from http://www.slideshare.net/ronreiter/writing-html5-web-apps-using-backbonejs-and-gae" instead.

11-Backbonejs/02-Backbone-Views.md
```diff
@@ -85,25 +78,27 @@ Rails views are awesome!  They are written in an HTMLish format with special han
 
 ### Underscorejs
```
Hamled: If we're not going to use the "js" suffix for Backbone, we should also not use it for Underscore.

11-Backbonejs/02-Backbone-Views.md
```diff
@@ -85,25 +78,27 @@ Rails views are awesome!  They are written in an HTMLish format with special han
...
+Notice the `<%= %>` handles.  Like in Rails, this will pre-process the result of the given expression or variable and put the result into the template.  You can identify this as a template by the script tag and the type attribute with the value of `text/template`.  
 
-Underscore supports the following kinds of templates:
--  `<%= %>` Interpolates the given expression into the surrounding HTML.  
-- `<% %>`   Executes the given Javascript expression 
-- `<%- %>`  Interpolates the value and makes it HTML escaped, meaning it allows the Javascript output as HTML.
+Underscore supports the following template functions:
+- `<%= %>` Interpolates the given expression into the surrounding HTML
+- `<% %>`   Executes the given Javascript expression
+- `<%- %>`  Interpolates the value and makes it HTML escaped, meaning it allows the Javascript output as HTML
```
Hamled: I'm not sure I understand what this is saying.

11-Backbonejs/03-Backbone-Models.md
```diff
@@ -49,7 +50,7 @@ So if the script above is run the console will result in:
 A new person has been instantiated.
 ```
 
-The defaults property lets you set default values to attributes for your model.  You can then retrieve attribute values with the `get` function.  
+The defaults property lets you set default values to **attributes** for your model.  You can then retrieve those  attribute values with the `get` function.  We add additional data to the object we pass in to `extend` by comma-separating those values.
```
Hamled: We should say "We can also provide other properties to the model by including them in the object we pass to `extend`."

11-Backbonejs/03-Backbone-Models.md
```diff
@@ -67,44 +68,52 @@ console.log("The person is named " + ada.get("name"));
...
 ```javascript
 var babbage = new Person({name: "Charles Babbage", age: 65, skills: "Hardware Design, Mathematics, Flower Arrangement."});
 ```
 
+These attributes do not necessarily need to have default values specified, or really be specified in any way. In the example above, we've provided two of the fields that we decided to set defaults for, and added another, `skills`.
+
 ## Get, Set & Unset
 
-Once you have an instance of your Model you can use get & set methods to set it's attributes.
+Once you have an instance of your Model you can use `get` & `set` methods to set it's attributes. You may pass individual keys with values, or you may pass an object containing any and all data that you might want to change.
```
Hamled: - "its" instead of "it's".
- We should probably refer to `set` and `get` as functions.
- We should say "For the `set` function you may pass two arguments, an attribute name and the attribute's value. Or you may pass an object containing multiple attributes to be set on the model."

11-Backbonejs/03-Backbone-Models.md
```diff
@@ -67,44 +68,52 @@ console.log("The person is named " + ada.get("name"));
...
 
-// prints undefined
-console.log(ada.get("skills"));
+console.log(ada.get("skills")); // undefined
 ```
 
 ## Adding Additional Methods to a Model
 
-You can add additional models to a method like other attributes:
+You can add additional methods to a Model like other attributes:
```
Hamled: We should probably say functions instead of methods, and properties instead of attributes.

11-Backbonejs/02-Backbone-Views.md
kariabancroft: Not sure - @CheezItMan did you pull this definition from somewhere?

droberts-ada:
Merging this in so that I have something to branch from.

PR #130, opened 2016-11-23
Expand the framework vs lib discussion

droberts-ada:
Inversion of Control!


kariabancroft:
Looks good! Thank you for expanding

PR #134, opened 2016-11-27
Bring the View lecture in line with the expected live code

droberts-ada:
There's a lot of content here. It may be this is the entire morning, and we don't get to events until after lunch. I think I would be OK with this so long as there are plenty of breaks.

It's also lacking in the "OK, now you go figure it out" department. I'm not quite sure how to fit this in - maybe @PilgrimMemoirs has some ideas?

### DIFF
--- a/11-Backbonejs/02-Backbone-Views.md
+++ b/11-Backbonejs/02-Backbone-Views.md
@@ -2,81 +2,155 @@
 # Intro to Backbone Views
 
 ## Learning Goals
-
 By the end of this lesson you should be able to:
 
--  Explain Backbone views and how to create them.
--  Be able to render a view as an underscore.js template.  
+- Explain what Backbone views are and why they're useful
+- Create a view around some data
+- Render HTML using a view
+- Use underscore.js templates to simplify and DRY up rendering
 
 ## What is a View?
+Backbone views are kind of middle-people in the Backbone world, filling a similar role to controllers in Rails. A view's job is to coordinate between the data and the DOM. When a DOM event happens, it's the controller's job to handle it and update the data as needed, and when the data changes it's the view's job to modify the DOM to match.
 
-Backbone views are kind of middle-people in the Backbone world.  They coordinate between your data, usually stored in Models and the DOM which displays them.  They also listen for events and respond to them as desired.  They can also plug into a Javascript templating library to determine how the information is displayed.  We will stick with Underscore.js' templating library.  
-
-## Creating a View
-
-Views can be created by extending Backbone.View.
+### Initial Setup
+Take a look at `src/app.js` in the project directory. You should see some sample data that looks like this:
 
 ```javascript
-var TodoView = Backbone.View.extend( {
-  initialize: function() {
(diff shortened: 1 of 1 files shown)


11-Backbonejs/02-Backbone-Views.md
```diff
@@ -2,145 +2,378 @@
...
 
-Drawing the view is done in the `render` function which determines how the view displays in the DOM.  Below the
+Views are created with the familiar `new` keyword. It will look something like this:
 
 ```javascript
-var TodoView = Backbone.View.extend({
-  initialize: function(){
-    this.title = "Study Ada Lovelace";
-    this.render();
+var taskView = new TaskView();
```
kariabancroft: Is this the naming convention we have followed? I'm not sure about using the upper and lowercase variables that are the same, but if this is what we're showing in the examples then this is good.

11-Backbonejs/02-Backbone-Views.md
```diff
@@ -2,145 +2,378 @@
...
+Chaining on `render()` is a very common thing in Backbone, and much of the code you'll find on the internet assumes you're using it. It will be much less painful to get into the habit of always returning `this` from `render()` now.
+
+#### Many Views
+One last step remains. Right now we're only viewing one task, but we want to see all of them. This can be solved with a quick loop:
+
+```javascript
+$(document).ready(function() {
+  var taskListElement = $('.task-list');
+  var taskViews = []
+  taskData.forEach(function(task) {
```
kariabancroft: should we use a different variable in the iteration variable to distinguish between this one and the one associated with the `task:` param in the initializer?

11-Backbonejs/02-Backbone-Views.md
```diff
@@ -2,145 +2,378 @@
...
-So how do Underscore templates play into this? 
+To add Underscore to your project, add the following line to the top of your `app.js` file:
+
+```javascript
+import _ from 'underscore';
+```
 
-Rendering raw html in the render function is both time consuming and painful to do.  Instead we can use a template attribute into our view.
+Without diving too deep into it, to use an underscore template you must do three things:
+1. Define the template (once)
```
kariabancroft: I think these need two spaces at the end to fix the formatting

kariabancroft:
I think this looks good. I agree with you that it will probably take the whole morning. Probably an hour on part 1: plain backbone views and an hour on part 2: underscore backbone views?

11-Backbonejs/02-Backbone-Views.md
droberts-ada: I see how this could be confusing. Here's an alternate:

`taskView` -> `card`
`taskListView` -> `application`

Not sure what we'll do about models, but we can cross that bridge when we come to it.

11-Backbonejs/02-Backbone-Views.md
droberts-ada: I think this will probably be fine as-is.

PR #135, opened 2016-11-28
Add lecture on handling DOM events in Backbone

droberts-ada:
I'm much more happy with the flow of this one. Since we handle two events, one simple and one complex, there's a clear moment to let them go figure something out for themselves. I think they'll need it after listening to us ramble on all morning.


kariabancroft:
Looks great! Nice job @droberts-ada

PR #145, opened 2016-11-29
Backbone Collections Lecture

droberts-ada:
I've moved Chris's details on Collection functions to another file. This stuff is valuable and should definitely be addressed, but I don't think we'll have time tomorrow. It also doesn't fit well with the live code, and would feel tacked-on if we left it there - it really deserves its own space. Perhaps for Wednesday?

I found a bug in the Models lecture, where we forgot to re-delegate events when the `TaskView`s were redrawn. I've added a section addressing this.

As with the Models lecture, this one is split in two. The first part is about retrofitting the app to use a collection, without adding any new functionality. In the second part, we add the ability to remove a task from the list.

Once they've gotten through this, they should be pretty much ready for Rolodex!

### DIFF
--- a/11-Backbonejs/04-Backbone-Models.md
+++ b/11-Backbonejs/04-Backbone-Models.md
@@ -283,6 +283,23 @@ Good catch, hypothetical student! Ideally we shouldn't be calling `render()` in 
 
 The reason is, if you're a view, you have no guarantee that all changes to the model will come through you. Maybe the model comes from a database server and is updated periodically. Maybe there are multiple views for the same model, and when one changes the others should pick it up. Model events can be an effective solution to all of these problems.
 
+### Re-Delegating DOM Events
+There's a bug in our application. Marking a task complete works fine, and adding a new task works fine, but once a new task has been added you can't mark _any_ task complete. Weird!
+
+The reason has to do with how DOM event listeners work. Remember that our `completeHandler` is bound to a click event on a specific button owned by our `TaskView`. But, look at `TaskListView.render()`. The first line reads
+
+```javascript
+this.listElement.empty();
+```
+
+That wipes out all the HTML in our task list, including all the buttons the `TaskView` click handlers are attached to! All the HTML is re-generated, but the click handlers are not.
+
+The solution is to rebind `TaskView`'s click handler, whenever a `TaskView` generates new HTML. In `TaskView.render()`, add the following line:
+
+```javascript
+this.delegateEvents();
+```
+
 #### Check-in Point
 Our new functionality should now be working. When the "Mark Complete" button is clicked, the title of the task gets struck out, and the button text changes to "Mark Incomplete". Your code should look [like this](https://gist.github.com/droberts-ada/0a18925644f6f1e5db1ec11eddf7dbba).
 
--- a/11-Backbonejs/05-Backbone-Collections.md
+++ b/11-Backbonejs/05-Backbone-Collections.md
@@ -135,9 +135,29 @@ var TaskListView = Backbone.View.extend({
 #### Check-in Point
 Right now, your application should have the same functionality as before: tasks can be created and marked complete. Your code should look [like this](https://gist.github.com/droberts-ada/a1e6ed27aea789d5fddf545c843058b6).
 
+## Deleting Tasks
+The next feature we will add to our task list app is the ability to a delete a task. This will give us exposure to how Collections interact with the Models they contain, as well as the types of events emitted by Collections.
+
+**Question:** Given what we've seen so far with Models, what code will we need to change to achieve this feature? What will the control flow look like in the app once it's done?
+
+### Setup
+First, we need to add a delete button. Since tasks should be deleted one by one, we'll add it to our template next to the "Mark Complete" button.
+
+```html
+<button class="button small-6 large-2 columns complete-button">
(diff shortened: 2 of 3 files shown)


11-Backbonejs/04-Backbone-Models.md
```diff
@@ -283,6 +283,23 @@ Good catch, hypothetical student! Ideally we shouldn't be calling `render()` in
 
 The reason is, if you're a view, you have no guarantee that all changes to the model will come through you. Maybe the model comes from a database server and is updated periodically. Maybe there are multiple views for the same model, and when one changes the others should pick it up. Model events can be an effective solution to all of these problems.
 
+### Re-Delegating DOM Events
+There's a bug in our application. Marking a task complete works fine, and adding a new task works fine, but once a new task has been added you can't mark _any_ task complete. Weird!
+
+The reason has to do with how DOM event listeners work. Remember that our `completeHandler` is bound to a click event on a specific button owned by our `TaskView`. But, look at `TaskListView.render()`. The first line reads
+
+```javascript
+this.listElement.empty();
```
Hamled: I believe we could also use jQuery's `detach` function instead of `empty` and this would keep the elements for our `TaskView`s around, with their event handlers still delegated.

The code for that would be something like:
```javascript
this.listElement.children().detach();
```

11-Backbonejs/05-Backbone-Collections.md
```diff
@@ -2,201 +2,262 @@
...
+## Introduction
 
-In Backbone a Collection is a type of Model that has other Model instances inside it.  You can think of it as similar to an Array, as an ordered collection of Models.  
+In Backbone a Collection is a type of Model that has other Model instances inside it.  You can think of it as similar to an Array: an ordered collection of Models.  
 
-Collections are useful for storing and manipulating a group of models.
+Using a Collection instead of an Array will bring us similar benefits to using a Model instead of a raw JavaScript Object. Collections can abstract away complex bits of logic (like only selecting Models that match a condition), they're really good at talking to APIs, and they emit events when they change.
 
-## Creating a Collection
+This lecture will be split into to parts, similar to the Models lecture. In the first part, we will retrofit our task list app to use a Collection instead of an array of Models, without adding any new functionality. In the second part, we will take advantage of this Collection to add a feature: the ability to delete a task.
```
Hamled: "split into to parts" -> "split into two parts"

11-Backbonejs/05-Backbone-Collections.md
```diff
@@ -2,201 +2,262 @@
...
-					new Person({name: "Cynthia", title: "Executive Director"})
-]);
-
-// Adding a Person via the add method.
-myPeople.add( new Person( {name: "Korica", title: "Program Coordinator"} ) );
+export default TaskList;
+```
+
+### Assembling the Collection
+In `app.js`, well create a new instance of our Collection from our raw task data, and pass it to our View. First, import our new `TaskList` constructor:
```
Hamled: "well create" -> "we'll create"

11-Backbonejs/05-Backbone-Collections.md
```diff
@@ -2,201 +2,262 @@
...
-
-    return this;
+    this.cardList = filteredList;
   }
 });
 ```
 
-And then in `app/js/app.js` we can create the collection and view and render it.
-
+Or more succinctly
```
Hamled: A couple of students have commented to me that they have found it very confusing when we've shown them how to do a thing one way and then immediately show them how to do it a different, but equivalent, way.

I think we should probably stick with the first of these two `removeTask()` functions, because it will probably be more familiar to the majority of students.

11-Backbonejs/04-Backbone-Models.md
Hamled: I think this should be `createTask` ? That's the function we're talking about in this section, at least.

The code in here does not match the code from the [check-in point diff](https://gist.github.com/droberts-ada/a1e6ed27aea789d5fddf545c843058b6) that is linked below.

Also, we should clean up the indentation in this code block.

Hamled: Not sure if it's worth bringing this in, but if we setup an event handler for the `add` event from our collection, which delegates to `addTask()` then we should be able to get away with:

```javascript
  createTask: function(event) {
    event.preventDefault();

    this.model.addTask(this.getInput());
    this.clearInput();
  }
```

11-Backbonejs/05-Backbone-Collections.md
CheezItMan: Agreed, it distracts from the main objective.

11-Backbonejs/04-Backbone-Models.md
droberts-ada: @Hamled In your second comment, what is the "this" that "should be `createTask: function(event) {`"?

Hamled: Sorry about that, I dunno why GitHub has pushed all of these comments together.

When I said `createTask: function(event) {` I was referring to [line 119 in 05-Backbone-Collections.md](https://github.com/Ada-Developers-Academy/textbook-curriculum/pull/145/files#diff-a5bbb6ccc9c373dcef0e90dc7b69a25dR119).

PR #147, opened 2016-11-30
Backbonejs

CheezItMan:


11-Backbonejs/04-Backbone-Models.md
```diff
@@ -291,6 +291,7 @@ The reason has to do with how DOM event listeners work. Remember that our `compl
 ```javascript
 this.listElement.empty();
 ```
+You will also need to update the TaskListView `getInput` function to use a Task model.
```
droberts-ada: I don't believe this is true. You should be able to `add` a raw JavaScript object to a collection, and Backbone will figure it out.

CheezItMan: Somehow had left that in from my earlier notes. fixed now

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,200 @@
...
+Below we have created a generic JavaScript object and then used Underscore to mix-in the Backbone.Events module into the object.  Then we add an event called, "Past_Due" to the object and Triggered the event with the `.trigger` function.  
+
+```javascript
+var task = {
+  assignee: "Dan",
+  task:     "Do work",
+  due_date: "01/01/2017"
+};
+
+_.extend(task, Backbone.Events);
```
droberts-ada: Brackets at least have not talked about `_.extend()`. Maybe expound this?

11-Backbonejs/xx-Event-Bus.md
```diff
@@ -0,0 +1,102 @@
...
+
+What did this do for us?  Now each of the TodoViews and the CurrentTaskView have a reference to the same bus object, which will let them communicate.  
+
+#### Adding an Event to the Bus
+
+Now we can create an Event in the Bus and let the TodoView trigger that event to publish information about that particular task.  
+
+In the CurrentTaskView we can specify an Event on the Bus that we want to listen to.  In the initialize method: 
+
+```javascript
```
droberts-ada: Whitespace

11-Backbonejs/xx-Event-Bus.md
```diff
@@ -0,0 +1,102 @@
...
+  	initialize: function(options) {
+    	this.bus = options.bus;
+    	var that = this;
```
droberts-ada: The that/this/context binding here is a little weird - probably deserves more explanation.

Can you use `listenTo()` in this context?

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+Backbone API Events can be added with the `.on` method as you saw earlier.  This lets you create custom events that can be triggered in your Models or Views to update other objects when something occurs, a change in status, user interaction etc.  
+
+Looking at the TodoViews in the Codepen, we can see when a task is clicked on using a DOM event.  However it would be REALLY useful if we could update another view when a Todo item is clicked on, for example to show more details on the selected Todo item.  Thus we return to the Backbone Events API and the idea of an Event Bus.
+
+### listenTo 
+
+Similar to the `.on` function, the `.listenTo` method lets a Backbone object listen to an event that occurs on another object.  Look below and see how `listenTo` differs from the `on` function.  The main difference between the two is the context in which they add callbacks.  `on` lets you directly add a callback function to another object.  `listenTo` tells an object to listen for events on another object.  This allows the 'listener' to keep track of the objects for which it is listening.
```
CheezItMan: I'll add more, plus when a view is destroyed `listenTo()` will automatically remove all event listeners, while on will not.  So you **can** end up with zombie listeners out there.

11-Backbonejs/06-Adding-Custom-Events.md
CheezItMan: Added more explanations about _.extend(), and the Event API.

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,200 @@
...
+
+1.  Setup an event handler for when a task card is clicked on.  
+1.  Then we can 'trigger' an event letting our TaskListView know that we want to edit this task.
+1.  When the task Lists' listener 'hears' the event it can delete the card being edited, and put the title & description fields in the input tags.
+
+There will be drawbacks, but we'll see how it works.
+
+### First create an Event listener in TaskView for when the `<div>` in the template is clicked on:
+
+```javascript
```
CheezItMan: Doh!  Added in

CheezItMan:
Ok, things are updated a bit.

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,257 @@
...
+
+var task = {
+  assignee: "Dan",
+  task:     "Do work",
+  due_date: "01/01/2017"
+};
+
+_.extend(task, Backbone.Events);
+
+task.on("PastDue", function(message) {
```
Hamled: We're listening for a `PastDue` event here, but triggering a `Past_Due` event. Also further on in  the paragraph under "Events: on, off & trigger" we use an event named `pastDue`.

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,257 @@
...
+
+task.trigger("Past_Due", {msg: "past due",
+							date: new Date()						
+});
+```
+
+You can play with this in action [here:](http://codepen.io/CheezItMan/pen/JbyrOG).
+
+
+When you fire an event with `.trigger`.  The function takes 2 parameters.  The first is the name of the event to trigger, while the second is an optional pargument to send to the function handling the event.  In the case above it was a simple message about the task being overdue.  Like above, when you need to pass multiple parameters you can simply pass a JavaScript object as the second parameter.
```
Hamled: "optional pargument" -> "optional argument"

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,257 @@
...
+  }
+});
+``` 
+
+If you wanted to listen for a specific item within the view you could, with jQuery, specify the item using a CSS Selector.  For example if there was a block within the view with a class `title` and you wanted to listen for clicks on that block, you could put this line into the events hash `"click .title": "onTitleClick"`.  This would cause the View to listen for clicks on any element within the view and with the `title` class.  When any sub-element with that class is clicked on the `onTitleClick` event handler will run.
+
+
+Backbone Events API allows you to add event handlers with the `.on` method as you saw earlier.  This lets you create custom events that can be triggered in your Models or Views to update other objects when something occurs, a change in status, user interaction etc.  
+
+Looking at the TodoViews in the Codepen, we can see when a task is clicked on using a DOM event.  However it would be REALLY useful if we could update another view when a Todo item is clicked on, for example to show more details on the selected Todo item.  Thus we will return to the Backbone Events API and the idea of an Event Bus later.
```
Hamled: I think in the Codepen we're using the name `TaskView` rather than `TodoView`. Also "Todo item" -> "Task item".

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,255 @@
...
+
+Here's another example:
+
+```javascript
+// A view listening to a Task Model's event.
+this.listenTo(this.task, 'change:completed', this.filterOne);
+this.task.on('change:completed', this.filterOne);
+```
+
+Notice that the `listenTo()` method has the listening object subscribe to another object's event.  The on method has the model, which generates the event remember to notify the callback function.  There's another difference between `listenTo()` and `on()`.  If you call another function `stopListening` any listeners added by `listenTo()` will be removed.  Similarly if a view is removed with the `remove` method all listeners using `listenTo()` will be removed.  Models behave the same way, models deleted with the `destroy` method will call `stopListening` on any subscribers using `listenTo()`  In the **old** days, the only way to subscribe to an event was with the `on()` function.  That caused problems because when a view or model was removed the callback functions were still listening for events, which wasted memory and lead to unexpected behaviors.
```
Hamled: * There's a whole lot going on in this paragraph, it'd probably be best to move the information about how/when to stop listening into its own paragraph.
* I'm not sure that the final couple of sentences about the old days is meaningful information for the students. If there's a technique to avoid those problems we should probably inform them of that technique explicitly.
* The second sentence "The on method has the model..." is confusing to me, I'm not sure what it's saying.

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,255 @@
...
+// in app/views/task_list_view
+
+// prints the view to the console
+this.model.on("update", function() {
+  console.log("this is:  ", this);
+}, this);  // <--- Note the "this" 3rd parameter.
+
+// prints the view to the console
+this.listenTo(this.model, "update", function() {
+  console.log("listenTo this = ", this);
```
Hamled: We should probably change this line to `console.log("this is:  ", this);` to maintain parity with the other examples and more readily show how calling `on` with a context of `this` has equivalent results as using `listenTo`.

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,255 @@
...
+
+// prints the model to the console. 
+this.model.on("update", function() {
+  console.log("this is:  ", this);
+});
+```
+
+Honestly it's something that will *almost* never be used.
+
+So... when in doubt use `listenTo()`.
```
Hamled: If we're going to suggest that they should use `listenTo` first, perhaps that should be the concept we introduce first with a smaller ending section about `on`. Or leave out the discussion of `on` entirely if we're not actually going to require it in their projects.

11-Backbonejs/06-Adding-Custom-Events.md
```diff
@@ -0,0 +1,259 @@
...
+## Check-in
+
+At this point your views should look like [this:](https://gist.github.com/CheezItMan/14346e3bcf1cb25879341713a849015d)
+
+
+#### Word Of Caution:
+
+Sometimes you may have a View listening for events on a particular Model.  If you remove the View in the course of your program without removing the corresponding Model and without turning off any event listening this will cause the View to hang around in memory as the Model still thinks it has to notify the 'ghost view' about events, even though the View has been removed from the the program.  This is a common form of 'memory leak' in Backbone.  If you use `on()` you will need to make darn sure you stop listening to events manually before removing any listeners.  
+
+A good reason to use `listenTo()`!
```
Hamled: As above, if we're going to strongly suggest that they use `listenTo` instead of `on` so as to avoid the problems mentioned above, which they likely do not understand at this point, maybe we should just not teach them to use `on`.

11-Backbonejs/06-Adding-Custom-Events.md
CheezItMan: True, but since I was originally linking this to jQuery events, on made a good parallel.  I'll bring `listenTo()` up.

droberts-ada:
Looks good to me.

PR #151, opened 2016-12-13
Rails Folder Structure

kariabancroft:


Hamled:
This looks fine to me, although I'm unclear on why the resources are being reordered. Is this based on how we had taught things in C6?

kariabancroft:
The numbering was completely wrong/duplicated/nonsensical previously. I'm reordering based on what we did in C6 so we can actually see what that is. And then we will be able to determine how/if we want to adjust for C7.

droberts-ada:
Do we need to do something similar for the intermediate rails folder? I remember there being a bunch of xx-whatever in there, don't remember if the order of everything else was right.

kariabancroft:
@droberts-ada Yes, we do. This is my starting point for now for this first folder

PR #160, opened 2017-01-02
Add pull request template

kariabancroft:
Once this file is added to our repository, it _should_ automatically populate any future pull requests with this markdown template. Are there other things that you think we should include in this template for ourselves to fill out when we make changes to textbook curriculum?

There is no real way to enforce that we fill this template out, but it will at least get us asking ourselves these questions when we submit changes.

This is also the type of thing that we can do with student repositories to provide them a different template for their self-feedback or just general submission guidelines for each project.

@Hamled @PilgrimMemoirs @droberts-ada @CheezItMan @sudocrystal

### DIFF
--- a/PULL_REQUEST_TEMPLATE
+++ b/PULL_REQUEST_TEMPLATE
@@ -7,5 +7,4 @@ Are any of these items still in progress?
 - [ ] Still need to add "Additional Resources"
 
 ## Feedback Requested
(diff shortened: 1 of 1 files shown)


CheezItMan:
Looks good.

kariabancroft:
@Hamled I think that's a good point. I will update the template to mention using the GitHub feedback stuff but then specifying details about the type of feedback the submitter wants.

PR #168, opened 2017-01-18
Add new project submission doc

kariabancroft:
## Description
This is a new resource to guide students in submitting their early assignments *without* the use of the terminal skills.

## Todos
N/A

## Feedback Requested
Is the level of detail appropriate for their level of understanding? This lecture will be happening pretty early, week 1 or 2 at the latest. I really mainly need Jamie or Dan to review/approve as one of ya'll will need to share this with your students.


PilgrimMemoirs:
If you want I can do something with the images to make them not disrupt the flow of the instructions as much.  If that still doesn't look quite right, I know taking a screenshot of the full viewport and adding circles around the areas to interact usually works well. 


For step 2 - even thought there is a photo, would be worth being explicit that they are only on their version on the repo if their name is at the top before the project name (like shown in the photo) and also in the url. 

It also might be worth splitting the instructions into two sections- one labeled more explicitly as adding a new files to github and the other is creating a pull request.  So really just adding headers above those two portions.

droberts-ada:
Looks good to me. I think the high level of detail will be important this early in.

PR #169, opened 2017-01-19
Expand vocabulary section in Project Submission doc

Hamled:
## Description
This adds a simple definition for the term repository (as we use it elsewhere within the document) with a couple of examples. It also adds an example to the definition of fork.

## Feedback Requested
Are these reasonable examples to use? Should we add an example of a pull request?

### DIFF
--- a/00-programming-fundamentals/06-submitting-projects-v1.md
+++ b/00-programming-fundamentals/06-submitting-projects-v1.md
@@ -5,7 +5,7 @@ We utilize GitHub for all of our curriculum and project submissions. Before we l
 ### Vocabulary
 **Repository**: The term that GitHub uses for a collection of files that are all part of the same project. Every repository on GitHub is owned by an individual (e.g. [Hamled/dotnvim](https://github.com/Hamled/dotnvim)) or an organization (e.g. [github/hub](https://github.com/github/hub)).
 
-**Fork**: A copy of the repository (right now, mostly just a `README` file) on your own personal GitHub account
+**Fork**: A copy of the repository (right now, mostly just a `README` file) on your own personal GitHub account. For example [AdaGold/Word-Guess](https://github.com/Ada-C7/Word-Guess) is the repository for the Word Guess project and [Hamled/Word-Guess](https://github.com/Hamled/Word-Guess) is Hamled's _fork_ of that repository.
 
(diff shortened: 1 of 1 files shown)


00-programming-fundamentals/06-submitting-projects-v1.md
```diff
@@ -3,7 +3,9 @@
 We utilize GitHub for all of our curriculum and project submissions. Before we learn how to use GitHub via the command line (terminal) we will need to be able to submit our projects using a process where we'll only use the web interface. Bookmark this page!
 
 ### Vocabulary
-**Fork**: A copy of the repository (right now, mostly just a `README` file) on your own personal GitHub account
+**Repository**: The term that GitHub uses for a collection of files that are all part of the same project. Every repository on GitHub is owned by an individual (e.g. [Hamled/dotnvim](https://github.com/Hamled/dotnvim)) or an organization (e.g. [github/hub](https://github.com/github/hub)).
+
+**Fork**: A copy of the repository (right now, mostly just a `README` file) on your own personal GitHub account. For example [AdaGold/Word-Guess](https://github.com/Ada-C7/Word-Guess) is the repository for the Word Guess project and [Hamled/Word-Guess](https://github.com/Hamled/Word-Guess) is Hamled's _fork_ of that repository.
```
kariabancroft: Did you link Ada Gold to Ada-C7 on purpose? I don't think that we'll create the C7 fork before the project is needed and so its quite possible that this link won't go anywhere

Hamled: I chose that one specifically because it does currently exist and I could create a fork of it.

Usually I create forks from AdaGold because that's were we push our changes, but I didn't want to potentially confuse the situation for new students since they shouldn't need to interact with AdaGold at all.

However I just noticed that I left the text as saying AdaGold/Word-Guess, so I will fix that.

Hamled: Actually, that's wrong. I originally was going to use the Ada-C7 one for the reasons stated above.

However I must have decided to use AdaGold anyways so that it would be more relevant to later cohorts? It seems the fork I created was directly from AdaGold, so I will change the link to reflect that.

PR #179, opened 2017-02-08
Updated getting-to-know-your-text-editor.md

CheezItMan:
## Description
I just toned down some of the attitude, took our packages we don't use (RSpec) and added one to autocomplete foundation elements.  I also added a definition of `linter`, as I didn't know the term when I came to Ada.   

I'm wondering if we should just drop sublime.  I'm not convinced it adds anything.

## Todos
- No

## Feedback Requested
Any feedback is helpful here.  If one person can look at it, that would be great.


kariabancroft:
Yeah - we definitely won't be using Sublime, but it's alright to leave in there for now I suppose

PR #188, opened 2017-02-21
class methods vs instance variables

kariabancroft:
## Description
I wasn't too pleased with either section of this curriculum resource so I refactored it a bunch. I tried to simplify the examples in the initial "review" section as well as provide some more context and verbiage around _why_ we're creating class methods.

## Todos
None

## Feedback Requested
1. I'd love a fine tooth comb on all changes. 
1. I'd love feedback on whether or not the `Pawn` example at the end is still useful. Primarily, it is doing the `self.new` inside of the class method, but I think it might be oddly confusing/complex. 
1. Looking for feedback on whether or not we should lecture using the `Pawn` example, or use it as an exercise to have folks review with their seat squad?

### DIFF
--- a/02-intermediate-ruby/03-class-methods-and-self.md
+++ b/02-intermediate-ruby/03-class-methods-and-self.md
@@ -172,7 +172,7 @@ puts Penguin.looks
 puts p.class.looks
 ```
 
-I n the second `puts` statement from above, we are utilizing the `class` on our instance to then call the **class method** because `p.class` will give us `Penguin`.
+In the second `puts` statement from above, we are utilizing the `class` on our instance to then call the **class method** because `p.class` will give us `Penguin`.
 
 ### Another Example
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/03-class-methods-and-self.md
```diff
@@ -57,36 +60,121 @@ order.to_money #=> $20.00
...
+puts p.name
+puts p.looks
+```
 
-## Class Methods & `self`
-Class methods are defined like instance methods, but they start with `def self.make_row` and are called internally with the `self` identifier.
+Which piece of this class will always be the same for **every instance** of the class?
 
-Class methods are called directly by the _class_ and not by an _instance_ of the class.
+#### Activity
```
kariabancroft: Hmmm that could be interesting...

PR #195, opened 2017-02-27
Fixed typo in "Nuances"

khyew:
## Description
O(2^N) is not O(N), so I'm guessing it should actually say O(2N) or O(2*N)


Hamled:
Thanks!

PR #199, opened 2017-03-06
Added let/before resource.

CheezItMan:
## Description
A short resource on testing with `let`.  

## Todos
None

## Feedback Requested
Any feedback is welcome.

### DIFF
--- a/02-intermediate-ruby/lazy-with-let.md
+++ b/02-intermediate-ruby/lazy-with-let.md
@@ -1,0 +1,180 @@
+# Being Lazy With Let
+## Learning Goals
+You should be able to: 
+- DRY up your code with let & before.
+- Explain how `let` is "lazy" in Minitest.
+
+In your Minitest code you will often find yourself creating instances over and over again in each test case like this:
+
+```ruby
+it "must have a name" do
+  player = Scrabble::Player.new("Ada Lovelace")
+  
+  # expectations go here
+end
+```
+
+In most test cases we're creating a new identical instance of `Player` to work with.  This doesn't seem DRY.  
+
+## Solution 1:  Before Blocks
+
+One solution is to use Minitest's `before` block. 
+
+```ruby
+describe Player do
+  before do
+    @player = Scrabble::Player.new("Ada Lovelace")
+  end
+
+  it "must respond to name" do
+    @dplauer.must_respond_to :name
+  end
+  ...
+end
+```
+
+So in the above example the `before` block runs before each test and provides an instance variable `@player` which we can use in our tests.  The problem with before blocks is that they **always** run before each test case.  Even if we don't want to use `@player` in a few test cases, like for instance if we want to try creating a Player with no arguments the before block will still run.  
+
+## Let
+
+Let is another way to DRY up your code.  With Let we create a block which can be referenced in our test cases.
+
+
+```ruby
+# player_spec.rb from Scrabble Project
+
+describe Player do
+  let (:player)  { 						 # <-- The name used
+  	Scrabble::Player.new("Ada Lovelace") # <-- The value returned by 'player'
+  }
+
+
+	it "has a name" do
+		player.name.must_equal "Watson"
+    end
+end
+```
+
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+
+	it "has a name" do
+		player.name.must_equal "Watson"
+    end
+end
+```
+
+The `let` statement creates a new method with the given name (in this case `player`) that executes the code in the curly braces and caches the result for later use.  That's why it's often called "lazy" the code inside the block doesn't execute until the let is first used, in this case when we do `player.name.must_equal "Watson"`.  So the let runs for each test case that uses it and doesn't run when it isn't needed.  Neat!
```
kariabancroft: I think these two sentences contradict each other b/c the first one says "executes". I don't think the first sentence is true.

02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Proof that let is lazy
+
+In this example using `before` the statement "Creating a player named Ada Lovelace" prints twice times showing that before always runs.
```
kariabancroft: Grammar issue with "twice times"

kariabancroft: I might clarify that there are two tests (aka `it` blocks) and that is why we see it twice

02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+### Proof that let is lazy
+
+In this example using `before` the statement "Creating a player named Ada Lovelace" prints twice times showing that before always runs.  
+
+```ruby
+# sample_spec.rb
+
+describe "Player" do
+  before do
+  	puts "Creating a player named Ada Lovelace"
```
kariabancroft: Indentation seems off here and below for some reason

02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+Creating a player named Ada Lovelace
+.Creating a player named Ada Lovelace
+.
+
+Finished in 0.011018s, 272.2817 runs/s, 272.2817 assertions/s.
+
+2 runs, 2 assertions, 0 failures, 0 errors, 0 skips
+```
+
+But if we use let instead:
```
kariabancroft: I think I'd be good to add headings above the `before` example and the `let` example so it's a bit easier to follow along with these sections

02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+
+Notice the `puts` ran only once, only in the test-case where it was used.  So let can often be more efficient because it's only ran where needed.  Unlike life, in Minitest being lazy pays off!
+
+### A Longer Let Example:
+
+The below example creates 2 let statements for a player and a tilebag from Scrabble and uses them together in a batch of tests.  You can use the names defined in let like variable names.  
+
+```ruby
+describe "Player & Tilebag" do
+  let (:player)  {Scrabble::Player.new('Watson')}
```
kariabancroft: If they write this in the single line - I'd prefer that they have extra spaces on either side of the curly: `{ Scrabble::Player.new('Watson') }`

02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+      player.players_tiles.length.must_equal 7
+    end
+
+  end
+end
+```
+
+## Summary
+
+So `let` is a useful helper method to DRY up your testing code.  With it you can define a method that is run only when used with the result cached so you can use it like a local variable in your test cases.
```
kariabancroft: I'm not sure the students have the context to what it means for something to be cached. Also, I'm not sure I understand that to be true in how let is executed.

02-intermediate-ruby/lazy-with-let.md
```diff
@@ -0,0 +1,180 @@
...
+  end
+end
+```
+
+## Summary
+
+So `let` is a useful helper method to DRY up your testing code.  With it you can define a method that is run only when used with the result cached so you can use it like a local variable in your test cases.  
+
+
+## Resources
```
kariabancroft: 👍

PR #200, opened 2017-03-06
Controller testing

CheezItMan:
## Description
Updated Controller testing for Rails 5 and using spec-style testing.  

## Todos
Nothing on this, although an exercise for controller testing is next for me.

## Feedback Requested
Feedback is welcome.


08-rails/testing-controllers.md
```diff
@@ -1,178 +1,226 @@
...
 Let's refresh our memory with a controller action:
 ```ruby
 def show
-  @student = Student.find(params[:id])
+  @post = Post.find(params[:id])
 end
 ```
 
-In this case, we need to provide a key of `id` with a corresponding value. We specify this in a hash to correspond to `params` along with the verb and action.
+In this case, we need to provide a key of `id` with a corresponding value in the path helper.
```
Hamled: I think it would be more accurate to say that we pass the ID value as an argument to the path helper. Alternatively, we could change the code snippet below to say:
```ruby
get post_path(id: 1)
```
and then this line would be most accurate.

08-rails/testing-controllers.md
```diff
@@ -1,178 +1,226 @@
...
-  - No one is logged in
-  - If the action touches a Model that belongs to the user, then you also need to test when the wrong user is logged in
+## Rails Matchers
 
-That's not an exhaustive list, but it's a good starting point.
+|   Matcher	|   Sample	|
+|---	|---	|
+|   `must_respond_with`	|   `must_respond_with :success`	|
+|   `must_redirect_to`	|   `must_redirect_to root_path`	|
+|   `must_change`	|   `proc {delete post_path(posts(:one).id) }.must_change 'Post.count' -1`	|
```
Hamled: We should have a comma between `'Post.count'` and `-1`.

PR #203, opened 2017-03-11
Dpr/css

droberts-ada:
## Description
Clean up whitespace and expand examples on some of the CSS lectures.

## Todos
N/A

## Feedback Requested
Copy editing

### DIFF
--- a/05-html-css/06-css-selectors.md
+++ b/05-html-css/06-css-selectors.md
@@ -65,20 +65,6 @@ h1::selection {
 
 **Question:** What is the difference between a pseudo-element and a pseudo-class? Why are they different things at all?
 
-## Multiple Selectors
-The same rule set can be applied to two different selectors by putting a comma in between selectors. Doing so can help keep your CSS stay DRY.
-
-Examples:
-
-```css
-h1, h2, h3  {
-  color: teal;
-  font-family: helvetica;
-}
-```
-
-*Any* combination of selectors can be used! For example:
-
 ## Selectors Based on Relationships
 Remember how we emphasized elements relationship with the sectioning elements lesson? Knowing relationships between elements gives us even more ways of selecting elements!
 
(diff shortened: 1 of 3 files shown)


05-html-css/06-css-selectors.md
```diff
@@ -39,67 +41,90 @@ Attribute selectors can get more specific, read [this article](https://css-trick
...
 ```
 
+**Question:** What is the difference between a pseudo-element and a pseudo-class? Why are they different things at all?
+
+## Selectors Based on Relationships
+Remember how we emphasized elements relationship with the sectioning elements lesson? Knowing relationships between elements gives us even more ways of selecting elements!
+
+Here are common selectors based on relationships
+
+| Selector        | Selects
```
kariabancroft: Could we add concrete examples to this table? Or in a table below?

05-html-css/06-css-selectors.md
```diff
@@ -39,67 +41,90 @@ Attribute selectors can get more specific, read [this article](https://css-trick
...
+```
 
 ## Practice!
 [CSS Diner](http://flukeout.github.io/)
 
 ## Vocab ✅
-  - Pseudo-Classe
-  - Pseudo-Element
-
+- Pseudo-Classe
```
kariabancroft: "class" not "classe"

05-html-css/06-css-selectors.md
droberts-ada: I don't know that it's really needed here - other than `A E`, I would consider these as "good to know it exists but you'll never actually use it". Especially since any examples will need both HTML and CSS, I think that anything we add will be a bunch of extra clutter, and not add a lot to student understanding.

droberts-ada: I don't know that it's really needed here - other than `A E`, I would consider these as "good to know it exists but you'll never actually use it". Especially since any examples will need both HTML and CSS, I think that anything we add will be a bunch of extra clutter, and not add a lot to student understanding.

PR #207, opened 2017-03-15
Add discussion of page flow to the layout-display resource

droberts-ada:
## Description
How does the browser decide where to put things? Using the flow! Gotta know about it - gives context for everything else.

Sorry it's a bit 11th-hour, hopefully there's time to look it over.

## Todos
N/A

## Feedback Requested
Too verbose? Are the examples actually informative? I basically just made up some terminology ("rendering element", "containing element") - are there real words for these things?


droberts-ada:
I did double check that all the images &c render properly on GitHub.

PR #209, opened 2017-03-16
Rails5

CheezItMan:
Just making sure that rake is replaced with rails where necessary.


kariabancroft:
Looks good! I'm making some updates to some of these files so i'm going to merge yours in now

PR #211, opened 2017-03-19
Initial document on how to switch to postgres

CheezItMan:
## Description
A document on how to setup rails with Postgres.  I'm uncertain as to the order this would come in.

## Todos
Things to do:
- [ ] Figure out where we would do this (before Farmar?).

## Feedback Requested
Please let me know any feedback.

### DIFF
--- a/08-rails/how-to-use-postgres.md
+++ b/08-rails/how-to-use-postgres.md
@@ -31,7 +31,6 @@ $ rails db:migrate
 ...
 ```
 
-
 ## Converting Rails Apps Using Sqlite3
 
 
@@ -73,9 +72,6 @@ test:
 production:
   <<: *default
   database: db/production.sqlite3
-
-
-
 ```
 
 to:
--- a/08-rails/how-to-use-postgres.md
+++ b/08-rails/how-to-use-postgres.md
@@ -3,9 +3,19 @@
 
 Postgres (or PostgreSQL) is an open-source database which can provide scalability to your application.  By default Rails development sites use sqlite3 which can work fine, but it can be easier to use the same database application in development & production.  
 
-## When SQLite is a good choice
+## SQLite vs Postgres
 
-SQLite is a "serverless" database which means there isn't a separate database program running to manage access to the data, instead the application uses the SQLite file 
+SQLite is a "serverless" database which means there isn't a separate database program running to manage access to the data, instead the application accesses an SQLite file directly, similar to how our previous Ruby programs accessed CSV files.  
+
+The advantage is that there's almost no configuration involved in setting up an SQLite database and no access-control setup.  So SQLite is simple to get up and running.  
+
+SQLite runs into trouble when multiple programs access the same database directly.  There's not a server coordinating access and keeping two users from writing to the same data.  Instead SQLite uses file blocks to keep one program from accessing the data while another is writing to it.  This is far less efficient than using a server to coordinate access.  So an SQLite database is not well suited to a site with lots of simultaneous accesses.  It can however serve for sites which receive low levels of traffic and which do not share a database with other applications.  For the most-part we do not use SQLite for a production website.  However it is a very easy-to-use database for development.
+
+Postgres is a more traditional client-server database which runs a separate process that Rails will communicate with to access data.  
+
+The downside is that we need to setup a Postgres server and deal with some configuration of the database before we can use it.  
+
+On the upside using Postgres on both the development and production ends means reduces any chances of problems when deploying our app on the web via AWS or Heroku. 
 
 
 ## Installing Postgres
@@ -57,6 +67,8 @@ gem 'pg'
 
 ### Editing the `config/database.yml` File
 
+The database.yml file specifies how the Database is configured for the Rails application.  In the below example you can replace <APPLICATION_NAME> with the name of the database you choose.  For example in a database of books, you could name it.  `database:  library_development` etc.  
+
 Convert `config/database.yml` from:
 
 ```ruby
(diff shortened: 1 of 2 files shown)


08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,106 @@
...
+test:
+  <<: *default
+  database: db/test.sqlite3
+
+production:
+  <<: *default
+  database: db/production.sqlite3
+
+
+
```
droberts-ada: Whitespace?

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,106 @@
...
+default: &default
+  adapter: postgresql
+  encoding: unicode
+  # For details on connection pooling, see rails configuration guide
+  # http://guides.rubyonrails.org/configuring.html#database-pooling
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+
+development:
+  <<: *default
+  database: postgres-app2_development
```
droberts-ada: Are these strings starting with `postgres-app2` dependent on the application name?

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,106 @@
...
+# ...
+test:
+  <<: *default
+  database: postgres-app2_test
+
+production:
+  <<: *default
+  database: postgres-app2_production
+  username: postgres-app2
+  password: <%= ENV['POSTGRES-APP2_DATABASE_PASSWORD'] %>
```
droberts-ada: Should we talk here about putting your database password in your ENV?

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,106 @@
+
+# Setting Up A Rails Project with Postgres
+
+Postgres (or PostgreSQL) is an open-source database which can provide scalability to your application.  By default Rails development sites use sqlite3 which can work fine, but it can be easier to use the same database application in development & production.  
+
```
droberts-ada: Is it worth talking about what the differences are between PostgreSQL and SQLite, and why you would want to run one over the other?

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,106 @@
...
+```bash
+$ rails new my-rails-project -d postgresql
+...
+$ rails db:reset
+$ rails db:migrate
+...
+```
+
+
+## Converting Rails Apps Using Sqlite3
```
droberts-ada: Could be converting an old app is enough trouble / complication to not be worth covering.

08-rails/how-to-use-postgres.md
CheezItMan: Definitely worth adding, I've added a bit here.

CheezItMan:
I'm not adding ENV material because I want to wait on Environment variables until we start talking about `dotenv`.  Not too much too early.

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,124 @@
...
+
+## SQLite vs Postgres
+
+SQLite is a "serverless" database which means there isn't a separate database program running to manage access to the data, instead the application accesses an SQLite file directly, similar to how our previous Ruby programs accessed CSV files.  
+
+The advantage is that there's almost no configuration involved in setting up an SQLite database and no access-control setup.  So SQLite is simple to get up and running.  
+
+SQLite runs into trouble when multiple programs access the same database directly.  There's not a server coordinating access and keeping two users from writing to the same data.  Instead SQLite uses file blocks to keep one program from accessing the data while another is writing to it.  This is far less efficient than using a server to coordinate access.  So an SQLite database is not well suited to a site with lots of simultaneous accesses.  It can however serve for sites which receive low levels of traffic and which do not share a database with other applications.  For the most-part we do not use SQLite for a production website.  However it is a very easy-to-use database for development.
```
Hamled: We could get more explicit here and say that SQLite is not intended for scenarios where the database exists on a separate computer from the application accessing it, which is the only setup supported by our chosen deployment services.

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,124 @@
...
+To install Postgres you can use the Brew package manager, then set brew to start postgres to launch every time the machine reboots and lastly install the pg gem to allow Ruby to talk to the database.  
+
+These steps you only need perform once.
+
+```bash
+$ brew install postgresql
+...
+$ brew services start postgresql
+...
+$ gem install pg
```
Hamled: I'm not sure this is necessary as they'll need to include the pg gem in their project Gemfiles as well, and if they're using RVM gemsets (which we either provide or instruct them to create in most projects), they won't end up using this global copy of the gem at all.

08-rails/how-to-use-postgres.md
```diff
@@ -0,0 +1,124 @@
...
+...
+```
+
+## To setup a rails project with Postgres
+
+After Postgres is installed and configured you can create new rails apps with postgres this way.  `rails db:migrate` is an essential step because it sets up the required tables in the database for a successful start of the rails server.  
+
+```bash
+$ rails new my-rails-project -d postgresql
```
Hamled: If we instruct them to create a `.railsrc` file in their home directory with `-d postgresql` in it, they won't need to add it every time they use `rails new`:

```bash
echo "-d postgresql" > ~/.railsrc
```

Hamled:
I do think it would be nice to have this setup as early as possible.

However since they're going to start the baseline for TaskList on Monday afternoon if we're going to have use it for that project as well they'll all have to complete at least the installation instructions by then which seems unlikely.

08-rails/how-to-use-postgres.md
CheezItMan: I added a bit on the end about separate database & application servers.

CheezItMan:
We can talk about running this through on Monday or not.

PR #213, opened 2017-03-22
Fix Markdown headers

Hamled:
## Description
Recently GitHub decided to [create a formal spec](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown) for GitHub Flavored Markdown.

One side effect of this, and the resulting updates to their Markdown parser & renderer, is that specifying a header without a space or new line following the hash characters will [create a paragraph instead](https://github.github.com/gfm/#example-34).

Apparently this was always invalid, but GitHub's earlier implementation allowed it so they decided to break backwards compatibility. As a result, I've updated our textbook repository in all instances where we used the invalid syntax.

## Todos
- [ ] Update Jump Start
- [ ] Update project repos
- [ ] Update misc. other repos

## Feedback Requested
I found the updated lines with automated searching, and changed them with automated substitution regular expressions.

If you find that you've an abundance of free time and very little direction for how to spend it, perhaps try gardening? Failing that, you could check through all of our curriculum manually to see if I missed anything.

### DIFF
--- a/00-programming-fundamentals/xx-debug.md
+++ b/00-programming-fundamentals/xx-debug.md
@@ -1,4 +1,4 @@
-#Bug Off!
+# Bug Off!
 ## Debugging with Pry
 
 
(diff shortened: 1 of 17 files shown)


droberts-ada:
I helped with the regex - almost as good as a real review

PR #219, opened 2017-04-05
Split up Rails testing resources

droberts-ada:
## Description
The model testing resource has been split into a general intro, specific setup instructions, and actual info about model testing.

To the setup instructions, which cover making your Rails project use minitest-spec, I added a Rails template file that will make all the steps we cover happen automatically whenever a new Rails app is created. Hopefully this will save us lots of time and pain later. That template lives at https://gist.github.com/droberts-ada/26a3bff1510b0b45fa5dd764a833483c

## Todos
- [ ] Go through the model testing resource and make sure the examples are useful and relevant. I remember this being particularly painful for C6.

## Feedback Requested
I tested the rails template pretty thoroughly, but if someone could go through and double check that it works on someone else's machine I would appreciate it.


08-rails/test-02-setup.md
```diff
@@ -0,0 +1,97 @@
...
+
+- At the end of the output from `rails new`, you should see something like
+    ```
+    generate  minitest:install
+    Running via Spring preloader in process 51275
+    force  test/test_helper.rb
+    insert  test/test_helper.rb
+    ```
+- Our two testing gems, `minitest-rails` and `minitest-reporters`, should be included in the `Gemfile`
+- Code matching what's above should have been inserted into `config/application.rb` and `test/test_helper.rb`
```
kariabancroft: Do you have a link that helped you create the template structure? It might be nice to just add an "Additional Resources" if folks are interested in exploring the documentation on this more

CheezItMan:
I've tested the template and it's working, I've been looking for something like that for a while!

Genius!

08-rails/test-02-setup.md
CheezItMan: https://www.sitepoint.com/rails-application-templates-real-world/

droberts-ada:
All the instructions were already present in the models testing resource, so they're not really new, but I will still go through and make the requested changes.

PR #221, opened 2017-04-08
Kab/oauth

kariabancroft:
## Description
Add independent resources around `flash` and `session` that don't directly relate to our OAuth curriculum. I included step-by-step instructions for what we'll need to do to get the `session` set up, but I anticipate a lot of the comprehension may come from answering and discussing these questions.

## Todos
- [] Remove anything from the original OAuth lecture?

## Feedback Requested
Does this cover the idea of using `session` independently?


09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+Let's see it! We are going to create a special form which is going to allow our authors to "log in". This is not a secure way to simulate logging in, but we are using this as our first stepping stone to creating "real" authentication, which we'll do later on. We'll allow the user to type an authors name which will "log in" as that author.
+
+We'll need to a create a few new things in our application in order to get set up for this:
+1. Create a login form and form submission route (think of how we do `new` and `create`)
+2. Create a special controller for managing this login process called `SessionsController`
+3. Add empty controller actions for the routes in step #1
+4. Create a form view where our user will enter in the author's name
+5. Create our initial controller action which will set the `session`
+```ruby
+# app/controllers/sessions_controller.rb
```
droberts-ada: I like this example. Do you have a books repo sitting around that has an author model added to it?

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+    if author
+      session[:author_id] = author.id
+      flash[:success] = "#{ author.name } is successfully logged in"
+      redirect_to root_path
+    end
+  end
+```
+
+**Question**: Now where can we utilize this logged in Author information?  
+**Question**: How you think you'd use the `session` in a logout action?
```
droberts-ada: Do we need to think about adding a new Author? Or is that handled by existing CRUD stuff?

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Key Takeaways
+Rails provides a few special hash-like objects that allow us to go above and beyond local and instance variables in our Rails applications.
+
+We will utilize `flash` and `flash.now` to provide additional feedback to our users as they use our applications.
+
+We will utilize `session` to keep track of logged in user information.
```
droberts-ada: Is it worthwhile making the comparison to `params`? It's another hash-like object provided by Rails, but it operates in a totally different way.

droberts-ada: Maybe we could make a table!

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+
+As with all variables that we set in our controller actions, the way we get this data to display to our users is to put it in a view! Since we can use `flash` in many different contexts, we will add this to our `application.html.erb` view so that any controller action will be able to use this in the future (though you can add this to any view you'd like).
+
+```html
+# app/views/layouts/application.html.erb
+...
+<body>
+  <% flash.each do |name, message| %>
+    <div class="<%= name %>"><%= message %></div>
+  <% end %>
```
droberts-ada: Put this in a section! Reinforce that semantic HTML!

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+
+  <%= yield %>
+</body>
+```
+
+This view code will also dynamically add a `class` to the message `div` so that we can add any distinguishing styles to these flash messages.
```
droberts-ada: Do we want to show this? Maybe just do it in class, not needed in the textbook.

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+
+Now let's see what we end up with!
+
+## `flash.now`
+
+The above example works because right after we set the `flash` value in the controller, we are doing a `redirect_to`. If we want to use a `flash` but with a `render` instead of a `redirect_to` then we need to use `flash.now` instead. The different usage is solely based on what comes next. `flash.now` will only stick around going to a `render` and `flash` will stick around going to a `redirect`.
+
```
droberts-ada: Do you think it's worth showing a more complex example for an error case?

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,117 @@
...
+</body>
+```
+
+This view code will also dynamically add a `class` to the message `div` so that we can add any distinguishing styles to these flash messages.
+
+Now let's see what we end up with!
+
+## `flash.now`
+
+The above example works because right after we set the `flash` value in the controller, we are doing a `redirect_to`. If we want to use a `flash` but with a `render` instead of a `redirect_to` then we need to use `flash.now` instead. The different usage is solely based on what comes next. `flash.now` will only stick around going to a `render` and `flash` will stick around going to a `redirect`.
```
droberts-ada: From testing it out right now, it seems like anything put into `flash` will be available in both this and the next controller action, whereas `flash.now` is available in just this controller action. I don't think that's quite what's reflected here.

droberts-ada:
The last comment is a question about `flash.now`, which for some reason github is marking as outdated. It isn't.

09-intermediate-rails/session.md
kariabancroft: Yes (if there is time) but it's not one of the learning goals

09-intermediate-rails/session.md
kariabancroft: Yes - the Stacks one has it https://github.com/Ada-C7/stacks-book-app

09-intermediate-rails/session.md
kariabancroft: They were just manually added up until this point (which is probably fine)

09-intermediate-rails/session.md
kariabancroft: True - It's confusing either way so I was trying to do a 1-to-1 comparison, even though flash will work for either technically, but not in a logical way on a render.

09-intermediate-rails/session.md
kariabancroft: Yes, possibly. Just didn't think of anything worthwhile. Model save issues should use the errors hash so I wasn't sure what other error scenarios we'd want to simulate.

09-intermediate-rails/session.md
kariabancroft: yeah it honestly might not be I'm not sure.... I don't know what categoreis are close enough to even create a table to describe similarities/differences

kariabancroft:
Thanks @droberts-ada for the comments and for the updates!

09-intermediate-rails/session.md
```diff
@@ -0,0 +1,129 @@
...
+</body>
+```
+
+This view code will also dynamically add a `class` to the message `div` so that we can add any distinguishing styles to these flash messages.
+
+Now let's see what we end up with!
+
+### `flash.now`
+
+The above example works because right after we set the `flash` value in the controller, we are doing a `redirect_to`. If we want to use a `flash` but with a `render` instead of a `redirect_to` then we need to use `flash.now` instead. The different usage is solely based on what comes next. `flash.now` will only stick around going to a `render` and `flash` will stick around going to a `redirect`.
```
CheezItMan: Maybe change:
`flash.now` will only stick around going to a `render` and `flash` will stick around going to a `redirect`.
to
`flash.now` will display the message on the current view while `flash` will persist to the next action (which redirect triggers).

09-intermediate-rails/session.md
CheezItMan: I don't think they're quite similar enough.

PR #229, opened 2017-04-16
Fix broken headings in Markdown files

bryant1410:
GitHub changed the way Markdown headings are parsed, so this change fixes it.

See [bryant1410/readmesfix](https://github.com/bryant1410/readmesfix) for more information.

Tackles bryant1410/readmesfix#1


Hamled:
Closing as this was already fixed by #213.

@bryant1410 It looks like your script may be getting some false positives about repos that still need to be fixed due to line-endings issues.

The file in this PR (for whatever reason) appears to have DOS line-endings and the fixed version your script generated has Unix endings, thus it did generate a diff even though the content wasn't changed.

bryant1410:
Yeah, opened an issue in bryant1410/readmesfix#4 regarding this. Sorry and thanks!

PR #234, opened 2017-04-26
Updates to OAuth Testing Resource

droberts-ada:
## Description
Removed all that junk about keeping track of a session in the test! Turns out you don't need it.

Substantially modified the way we interact with the mock data. We now set it explicitly before every login request, and the `login` helper function now takes a `User` object to log in as. That should enable some nice testing around multi-user interaction, should students find the time. Also, turns out the request was completely ignoring the `env` argument, so I've removed it for simplicity.

Added `What Did We Accomplish` section at the bottom, because this is a big sucker and I'll probably have forgotten what all we did by the time I'm done teaching it.

### Warning
All the code snippets I used, particularly those interacting with the model, reflect the books oauth strategy and migrations and such we did in the Queues classroom. YMMV if you copy/paste.

## Todos
N/A

## Feedback Requested
Anything you've got time for.

### DIFF
--- a/09-intermediate-rails/testing-auth.md
+++ b/09-intermediate-rails/testing-auth.md
@@ -28,7 +28,9 @@ Because our integration tests are going to run code from all areas of our Rails 
 
 Specifically, we need to write our integration tests _in the role of the **browser**_. That is, our integration tests simulate a browser accessing our site by making HTTP requests to the server. Just like the browser, our test code is limited to making those requests and asserting various things about the response received back.
 
-In particular, this means we cannot modify the `session` directly from our tests. Instead we do what the browser would do: send another request.
+In particular, this means we cannot modify the `session` or `flash` directly from our tests. Instead we do what the browser would do: send another request.
+
+Note that we can *read* `session` and `flash` *after* a request has been sent, to check they were filled in correctly. It is only attempting to *set* them from a controller test that Rails disallows. This will be important later when we're checking error messages.
 
 ## Integration Tests and OAuth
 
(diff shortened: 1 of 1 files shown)


kariabancroft:
I think these updates are fine but just because we _can_ test it this was doesn't mean we _should_. I specifically disagree with modifying the User model to include any mocking data because that crosses the "changing prod code to get tests to work" line for me. I think the conversation of controller testing in rails should be a larger one for us (because I think there is some interesting data out there about whether or not this is worthwhile). I think asserting on the session makes sense but it doesn't really make sense why we'd have access to the session in the greater context. You can also directly set an auth_hash using Rails.application.env_config but I avoided this path (even though it's suggested by OmniAuth docs) because it's too egregious in that it's leveraging existing functionality in sort of hacky ways. I think this approach will work but I definitely think the mock auth logic needs to be in the test code not the model.

kariabancroft:
Also, just to clarify, I think the other solution also has issues. I like that your option wouldn't really require modifying all of existing student tests.

09-intermediate-rails/testing-auth.md
```diff
@@ -52,151 +73,284 @@ class ActiveSupport::TestCase
...
-        'omniauth.auth' => OmniAuth.config.mock_auth[:github] # using the configuration we added above
+# app/models/user.rb
+class User < ApplicationRecord
+  # ...
+  # Whatever was here before (relations, validations, etc.)
+  # ...
+
+  # Test helper method to generate a mock auth hash
+  # for fixture data
+  def mock_auth_hash
```
CheezItMan: It just seems off to me to modify the model with test-code.  Test code, I feel should go into our testing files, not the models or controllers.

Maybe instead in the test_helper file.

```ruby
def mock_auth_hash(user)
  return {
    email: user.email,
    uid = user.oauth_id,
    info: {
      email: user.email,
      nickname: user.username
    }
  }
end
```

09-intermediate-rails/testing-auth.md
```diff
@@ -52,151 +73,284 @@ class ActiveSupport::TestCase
...
-    ...
+  describe "Guest users" do
+    # we allow only the book index page for our guest users
+    # so we'll want to verify the redirect to root and message for these
   end
+end
 ```
 
-Now, within our existing tests, we'll want to use the `@session` variable to reference our session. This will ensure that we're testing the full integration now with the logged-in user.
+#### Logged In Users
```
CheezItMan: I like how this helps us avoid having to rewrite existing tests.

PR #242, opened 2017-05-15
Clarify examples on the javascript objects lecture

droberts-ada:
## Description
The examples were really weird, and didn't match up to what inheritance is actually used for. Hopefully they are now a little clearer.

## Todos
N/A

## Feedback Requested
- Do we want to actually talk about inheritance?
- These examples match the way I think about objects, but it might be there's something in how it was before that I'm missing.
- Do I have semicolons everywhere?


10-JavaScript/03-objects.md
```diff
@@ -0,0 +1,133 @@
...
+Note the `new` keyword. The syntax is a little different from Ruby, and it's easy to get tripped up.
+
+Note also that `Dog.prototype` is just a JavaScript object. It happens to be the object that JavaScript looks for when a constructor is invoked with `new`, but other than that it behaves just as we've seen before. So you could do things like:
+
+```javascript
+// Call a function from the prototype directly
+Dog.prototype.speak();
+
+// Add a function to the prototype
+Dog.prototype.isHungry = function() {
```
kariabancroft: It might be worthwhile to continue this example of extending the prototype, but I think it would depend on the class. It doesn't seem like it is a major learning goal.

PR #256, opened 2017-07-06
Add resource on references vs values

droberts-ada:
## Description
In C7, I did an impromptu lecture on pass by reference vs pass by value, and it ended up being super valuable for the students. I have expanded this into a full textbook resource with more illustrative examples, diagrams and background material.

Target is a 1 hour lecture for the 3rd or 4th week of curriculum.

[Original lecture notes](https://gist.github.com/droberts-ada/5f36d0d4ee1064a07a16cc0ca78eb439)

## Todos
None

## Feedback Requested
- Review my code!
- I've tried to go into a useful level of depth about what's going on under the hood without being terrifying. Have I actually achieved this?
- Are these diagrams good enough? Are there enough of them?


susanev:
I looked at this super quickly, as we do cover a bit about this in JSL, I noticed the use of the word "equal", "If we assign one variable to equal another variable", I'd recommend removing that word entirely because it tends to trip students up, so how about "If we assign one variable to another variable". I'll look more closely this weekend and give more feedback if that makes sense.

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+The results:
+
+```
+Inside short_strings_1, input is
+["dog", "par", "cat", "lla"]
+After calling short_strings_1
+["dog", "parrot", "cat", "llama"]
+```
+
+Seems like our method is indeed creating a list of shortened words, but our outer variable isn't being updated. The reason why has to do with references and values, and how data is stored in a computer.
```
kariabancroft: Can you update this sentence to reference the specific variables used in the method and outside the method

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+
+As an aside: one way to fix our method is to simply return the new array, and when calling it say `pets = short_strings_1(pets)`. However, sometimes this isn't an option - for example, what if our method was supposed to return the number of words that had to be truncated?
```
kariabancroft: Let's remove the word "simply" here

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+
+## References and Values
+
+When we create an array in Ruby (or a string or a hash or any other complex data type), we're actually creating two things.
+
+The first is the _value_ of the array, which involves asking the operating system for a bit of memory and then putting our data in it. You can think of this as the actual object. Each piece of memory we get from the OS has an _address_ representing its physical location in hardware, which is how we get back to it later.
+
+The second is a _reference_ to the array, which ties together the address of that memory with a name for our program to use. References are sometimes called _pointers_ (especially in C), and we say that a variable _points to_ or _references_ an object.
+
+![references and variables](images/references_and_values/references_and_values.png)
```
kariabancroft: It's not too clear to me where the 1234 comes into play here (i mean, i know what it is but it doesn't seem clear in the diagram) - so it might be good to put obj ID in parens or something along those lines.  It might also be good to copy the pets array before the diagram so folks can recall what that variable was referring to

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+
+```ruby
+# Two variables can point to the same object
+repeat = veggies
+puts "repeat.object_id: #{repeat.object_id}" # same as veggie.object_id
+```
+
+![referencing a variable twice](images/references_and_values/two_pointers.png)
+
+If we make changes to the object through one variable, you can see the changes via the other. The variables are just names, but the underlying object is the same.
```
kariabancroft: Maybe "The variables have different names, but ...."

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+puts "#{veggies}"   # ["turnip", "onion", "potato"]
+```
+
+![creating a new array](images/references_and_values/new_array.png)
+
+So to summarize, if two variables point to the same underlying object:
+- Modifications to the _object_ (the value) will be visible from both variables
+- Reassigning one _variable_ (the reference) with `=` does not affect the other variable
+
+Note that `+=` and the other shorthand operators all involve reassignment. If we say `veggies += ['rutabaga']`, Ruby creates a new array, copies all the values from `veggies`, adds in `rutabaga`, and reassigns `veggies` to point to this new array. This is true of strings and numbers as well.
```
kariabancroft: I wonder if it might be good to highlight the difference between the `+=` and `<<` here? I can see that being a question that will come up since we tend to use `<<` more often in our curriculum

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+
+**Question:** Given what we've learned, how can we modify our `short_strings` method to do what we want?
+
+The answer is to modify the underlying object, rather than reassigning the parameter. Here's what the resulting code might look like:
+
+```ruby
+def short_strings_2(input)
+  input.each_with_index do |word, i|
+    # Slice characters 0 to 2
+    input[i] = word[0..2]
```
kariabancroft: The return value of this method is unexpectedly different than the previous version, so can we update to keep that consistent? I don't think changing that is one of our goals here

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,267 @@
...
+  str.concat(' modified')
+  puts "inside modify_string, str is '#{str}'"
+end
+
+text = 'original'
+modify_string(text)
+puts "outside modify_string, text is '#{text}'"
+```
+
+Primitive types like numbers, booleans and `nil` follow basically the same rules. The catch is there's no way to change the underlying value of a primitive without reassignment. In programming lingo, we say that these types are _immutable_. This means that whenever you change the value, Ruby makes a copy and changes that instead.
```
kariabancroft: Do we use the vocabulary referring to "primitive types" in other places? (honest question)

susanev: I was under the impression that Ruby did not have primitive types at all.

droberts-ada: As far as I'm aware, even though integers, floats, bools and nil are all "objects", the ruby interpreter represents them differently than "real" objects. The value is stored directly on the stack rather than in allocated memory. This means there's no "underlying object", and you can't have two references to the same integer.

That being said, "primitive" probably isn't the right word - see [this stackoverflow](https://stackoverflow.com/questions/18790442/are-there-primitive-types-in-ruby).

02-intermediate-ruby/references-and-values.md
droberts-ada: Is it different? I think both will return whatever's currently stored in `input`.

```ruby
# Version 1
def short_strings(input)
  result = []
  input.each do |word|
    # Slice characters 0 to 2
    result << word[0..2]
  end
  input = result   # implicit return of input / result
end

# Version 2
def short_strings(input)
  input.each_with_index do |word, i|
    # Slice characters 0 to 2
    input[i] = word[0..2]
  end    # implicit return of the result of `each`, that is, the (now modified) input array
end
```

02-intermediate-ruby/references-and-values.md
droberts-ada: Changing the diagrams is a bunch of work unfortunately. Do you think a quick "Assuming the pets array is stored in address `1234`" would suffice?

02-intermediate-ruby/references-and-values.md
kariabancroft: Hmmm... You're right. I was focusing on the return of `each_with_index` being the original array  and not the fact that the block is mutating it. Can you add the ruby comments you added to this comment to the textbook code?

02-intermediate-ruby/references-and-values.md
kariabancroft: I think we could try it. I'm just worried that it looks like it could be code and if folks think you are doing a hash-like key/value thing they are going to get really hung up on it.

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,279 @@
...
+
+```ruby
+pets = ['dog', 'parrot', 'cat', 'llama']
+```
+
+and imagining that the `pets` array has been stored at address `1234`, we would get the following memory layout:
+
+![references and variables](images/references_and_values/references_and_values.png)
+
+Every variable in Ruby consists of these two parts, a reference and a value. Normally when you type the variable's name, Ruby automatically goes and gets the object. If you want to find out the address, you can use the `object_id` method:
```
shrutivanw: I realize I'm being a bit pedantic - so feel free to push back. :) I'm struggling with saying that _object_id_ returns the _address_. I understand that for most objects, object_id in Ruby would be the same as VALUE that represents the object at the C level. But there are special cases for true, false, nil and FIXNUMs in Ruby (very clever special case, but special case none the less). 
Could we change this to something like; "If you want to find out which object you are working with, you can use the 'object_id' method:"

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,279 @@
...
+
+So to summarize, if two variables point to the same underlying object:
+- Modifications to the _object_ (the value) will be visible from both variables
+- Reassigning one _variable_ (the reference) with `=` does not affect the other variable
+
+#### Identifying Reassignment
+
+One subtle point is that `+=` and the other shorthand operators all involve reassignment. If we say `veggies += ['rutabaga']`, Ruby creates a new array, copies all the values from `veggies`, adds in `rutabaga`, and reassigns `veggies` to point to this new array. This is true of strings and numbers as well.
+
+On the other hand, `<<` does _not_ involve reassignment. `<<` is shorthand for the `.push()` method, which changes the underlying object, not the variable itself (most methods work this way). Saying `veggies << 'rutabaga'` will modify the original array pointed to by `veggies`, and other variables pointing at that array will be able to see the changes as well.
```
shrutivanw: Another pedantic note :); could we change from "...original array pointed to by `veggies`, and ..." to "...original array referenced by `veggies`, and ..."

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,279 @@
...
+
+As an aside: one way to fix our method is to return the new array, and when calling it say `pets = short_strings(pets)`. However, sometimes this isn't an option - for example, what if our method was supposed to return the number of words that had to be truncated?
+
+## References and Values
+
+When we create an array in Ruby (or a string or a hash or any other complex data type), we're actually creating two things.
+
+The first is the _value_ of the array, which involves asking the operating system for a bit of memory and then putting our data in it. You can think of this as the actual object. Each piece of memory we get from the OS has an _address_ representing its physical location in hardware, which is how we get back to it later.
+
+The second is a _reference_ to the array, which ties together the address of that memory with a name for our program to use. References are sometimes called _pointers_ (especially in C), and we say that a variable _points to_ or _references_ an object.
```
droberts-ada: I still think it's important to introduce the vocabulary, since it's something they're likely to hear / read around on the internet. What about something like "The address part of a reference is sometimes called a _pointer_ (especially in C), and..."

02-intermediate-ruby/references-and-values.md
droberts-ada: I like that new wording.

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,278 @@
...
+
+`short_strings` is indeed creating a list of shortened words and storing it under the name `input`, but our outer variable `pets` isn't being updated. The reason why has to do with references and values, and how data is stored in a computer.
+
+As an aside: one way to fix our method is to return the new array, and when calling it say `pets = short_strings(pets)`. However, sometimes this isn't an option - for example, what if our method was supposed to return the number of words that had to be truncated?
+
+## References and Values
+
+When we create an array in Ruby (or a string or a hash or any other complex data type), we're actually creating two things.
+
+The first is the _value_ of the array, which involves asking the operating system for a bit of memory and then putting our data in it. You can think of this as the actual object. Each piece of memory we get from the OS has an _address_ representing its physical location in hardware, which is how we get back to it later.
```
Hamled: I think we could simplify these two paragraphs by removing the details about memory allocation.

For example, we could say that the _value_ of the array is "the actual data that our computer stores somewhere" and a _reference_ to the array is the "location where the actual data can be found."

Divorcing the discussion of references and values from the actual implementation Ruby uses would also make it easier to provide examples of other places where we use references:

| Reference | Value |
|:-----------|:-------|
| URL | Web page or file |
| File path on hard drive | The contents of that file |
| ID number in CSV | Row of data with that ID in another CSV file |

This would, however, necessitate changing a number of the excellent diagrams you've created.

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+![referencing a variable twice](images/references_and_values/two_pointers.png)
+
+If we make changes to the object through one variable, we can see the changes via the other. The variables have different names, but the underlying object is the same.
+
+```ruby
+veggies[1] = "onion"
+repeat.push("potato")
+puts "#{veggies}"     # ["turnip", "onion", "potato"]
```
Hamled: I think this code sample would be more meaningful if we moved this line above the `repeat.push` line and used the comment to show that its value before `push` was `["turnip", "onion"]`.

It'd probably also be a good idea to further highlight, in text after the code sample, that both times we're `puts`ing the variable `veggies`, but we called `push` on the variable `repeat`.

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,278 @@
...
+  # = changes the reference
+  param = ["new", "array"]
+  puts "  after reassignment, param.object_id is #{param.object_id}"
+  puts "  with value #{param}"
+  puts "  Finish reassign_parameter"
+end
+
+pets = ["dog", "parrot", "cat", "llama"]
+puts "Before reassign_parameter"
+puts "pets.object_id is #{pets.object_id}"
```
Hamled: For clarity we should also show the value of `pets` at this point.

02-intermediate-ruby/references-and-values.md
```diff
@@ -0,0 +1,278 @@
...
+- Discuss how objects are stored in Ruby
+- Differentiate between _references_ and _values_
+- Compare _modifying an object_ with _reassigning a variable_
+
+## Motivation
+
+We're going to start today with some Ruby code that does something a little unexpected. It's a method that takes an array of strings as an argument, and truncates (chops off the end of) all the strings with more than three characters. Or at least that's the idea.
+
+```ruby
+def short_strings(input)
```
Hamled: Since we specifically mention `+=` versus `push` later on I think this example code could be rewritten to use those methods, and be simplified at the same time:

Original:
```ruby
def add_llama(pets)
  pets += ['llama']
end

pets = ['dog', 'parrot', 'cat']
add_llama(pets)
puts pets
```

Fixed:
```ruby
def add_llama(pets)
  pets.push 'llama'
end

pets = ['dog', 'parrot', 'cat']
add_llama(pets)
puts pets
```

(Or use `<<` instead of `push`, whichever the students would be more familiar with at this point.)

02-intermediate-ruby/references-and-values.md
droberts-ada: Ooooh I like that table, I might just steal it.

02-intermediate-ruby/references-and-values.md
droberts-ada: This code already uses `<<`

02-intermediate-ruby/references-and-values.md
droberts-ada: I think I'm going to leave the addresses in, but add a disclaimer that you'll almost never use it in Ruby. Mostly because I don't want to redo all my diagrams.

PR #260, opened 2017-07-12
Add more theory to the objects lesson

droberts-ada:
## Description
In a classic Dan move, while attempting to build a vocab list for this lesson I accidentally rewrote a bunch of it. I've tried to keep to the spirit of what came before, just adding some discussion of theory / motivation / context. Big changes include

- Talk about the idea that an object ties together _state_ and _behavior_. We return to this mantra throughout the lecture, evaluating a built in class (`String`) and adding it to our own (`Chair`).
- Try and give motivation and context for why we're using objects / what problem they solve, and how this might fit into a real program.
- Reorder the practical section a bit - the code samples are all basically the same.
- Add a name to the `attr_whatever` pieces: "Macros"
  - [I even asked on Stack Overflow](https://stackoverflow.com/questions/45044789/in-ruby-what-is-the-vocab-word-for-attr-accessor/45045909#45045909)
- Actually add a vocab list

## Todos
None

## Feedback Requested
I've added a lot of text - is it too much? To dry, or overwhelming for the first couple weeks?


01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -17,14 +18,61 @@ Array.new
 # => []
 ```
 
-Where we would normally use the _string literal_ `""`, we can also call the `.new` method on the `String` class to create a new string object. A class can be identified by it's __capitalized first letter__. So `""`, `[]`, and `{}` are syntactic sugar; using _literals_ to _initialize_ a new _instance_ of an _object type_. Most classes do **NOT** have this kind of convenience.
+A class can be identified by it's __capitalized first letter__. In the above snippet `String`, `Hash` and `Array` are all classes. When you call `.new` on a class, you get an _instance_ of that class. In any program, there may be zero, one or many instances of any given class.
+
+While we normally use _string literals_ like `""` or `"elephant"` to create a new string, we can also call the `.new` method on the `String` class to create a new string object. `""`, `[]`, and `{}` are syntactic sugar; using _literals_ to _initialize_ a new _instance_ of an _object type_. Most classes do **NOT** have this kind of convenience.
```
kariabancroft: Should we even share these examples since (a) we don't encourage them to use these classes this way and (b) they haven't seen them used this way?

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -17,14 +18,61 @@ Array.new
...
+
+### State and Behavior
```
kariabancroft: It's so hard for me to understand this paragraph of text. Could we possibly have a thing like:

- **state**: data or attributes the object will keep track of
   ex: the abc in a string "abc"
- **behavior**: methods that _rely upon_ or _modify_ the state
   ex: the String upcase method

```ruby
# state
my_animal = "elephant"
your_animal = "giraffe"

# behavior - same method but results vary based on state
my_animal.upcase # "ELEPHANT
your_animal.upcase # GIRAFFE
```

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -17,14 +18,61 @@ Array.new
...
+
+In Ruby, state will be represented using _attributes_, also known as _instance variables_, and behavior will be defined using _methods_.
+
+### What Are Objects Used For?
+
+It can be intimidating to look at `String`, `Hash` or `Array` when learning about objects for the first time. These are fundamental programming concepts that get used in almost every program, how can we ever hope to build or reason about something as complex as that?
+
+Although any abstract idea can be represented as an object, it's sometimes easy to think of a physical object to get the idea. Let's imagine a program designed to keep track of a library catalogue, which might have a class called `Book`. A book has _state_, like `title`, `author`, and `copies_checked_out`. Some of this changes frequently and some of it will never change, but it's still all state. A book in our catalogue also has _behavior_, like `check_in`, `check_out`, or `place_hold`.
```
kariabancroft: Can this be turned into a small brainstorm question so they have to think about this themselves?

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -17,14 +18,61 @@ Array.new
...
+Note that all this state and behavior is for a particular instance of a `Book`. Think about the abstract concept of a book, what is its title or author? That doesn't make any sense. Only when we create an instance do these properties come into focus.
+
+When modeling a real-world object like a book or a person, it's important to remember that you don't need to track everything about it, only what's important to your program. You certainly could include information about the color of the book's cover or how tall it is in your catalogue, but you should only do so if that information is useful to you.
+
+Different kinds of objects have different state and behaviors, and the same real-world object in two different programs might have totally different representations. Think about how different a model of a person would be in a medical application versus in a gradebook. What information is relevant is totally different for the two programs.
+
+## Classes in Ruby
+
+As with methods, there are two steps to working with classes in Ruby:
+1. Define the class using the `class` keyword
```
kariabancroft: Can you bold "define" and "create an instance" below?

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -63,50 +111,112 @@ Chair.new # => #<Chair:0x007fb7da550ca0> - this is an instance of a chair, it ha
...
+
+puts kari_chair.style     # => swivel
+puts charles_chair.style  # => comfy
+```
 
-With those _parameters_, we can create _instance variables_. By prefixing a variable name with an `@` symbol, we create a variable that has visibility within __any instance method of the class__. As long as that specific instance of the class is around, Ruby will retain the value of its _instance variables_. Let's alter our `Chair` _class_ to look like this:
+Methods that access instance variables don't have to be simple getters or setters - they can have complex logic like any other method. For example, we might have `@seat_height` and `@max_height` instance variables and an `adjust_seat` method:
+
+```ruby
+def adjust_seat(direction, amount)
```
kariabancroft: I really like the idea of ensuring that we can show that there are methods other than the getters/setters that we may want to have within the class. I feel like this example is way too complex and is using variables that haven't been defined or set anywhere else (`@seat_height`). How about a simple "print" function that returns a nicely formatted string with the values from the instance variables above?

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -4,6 +4,7 @@
   - Use our first _instance variables_
```
kariabancroft: I feel like based on the updates that have been made, we should really update these learning goals to be more specific

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -115,33 +225,64 @@ class Chair
...
+```ruby
+kari_chair = Chair.new('swivel', 'black')
+puts kari_chair.style     # => swivel
+```
+
+Much more concise! This pattern of passing in a bunch of values for instance variables is a very common one, but the constructor is just another method, so you can write any code you need in it. You can even call other methods!
+
+Another advantage of setting instance variables in the constructor is that we know those variables will always have a value. By making it impossible to have a `Chair` without an associated `style` and `color`, we can save ourselves all sorts of frustration later on.
+
+### Use Macros to Avoid Repetition
```
kariabancroft: Could we just call these helper methods?

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -115,33 +225,64 @@ class Chair
...
+
+The code above allows us to read/get and write/set the color and style properties in the `Chair` _class_. This is done so frequently that Ruby added some syntactic sugar to help us out. Enter two _macros_, `attr_reader` and `attr_writer`:
 
 ```ruby
 class Chair
-  attr_reader :name, :color
-  attr_writer :name, :color
+  attr_reader :style, :color
```
kariabancroft: Could we show it with only one attribute for at least one so they don't get the impression that you have to have them for every instance variable?

01-ruby-fundamentals/07-oo-classes.md
```diff
@@ -63,50 +111,112 @@ Chair.new # => #<Chair:0x007fb7da550ca0> - this is an instance of a chair, it ha
...
 ```
-__Question: What are the limitations of our `Chair` _class_?__
 
-### `initialize`
-Let's start over building our class, but let's build it to be a little customizable. We want to be able to customize the attributes `color` & `name` when we create our chair.
+### Adding State Through Instance Variables
+
+Ruby allows us to attach variables to an instance of a class. Such variables are appropriately named _instance variables_. To create or use an instance variable, prefix the variable's name with the `@` symbol. Instance variables can be read and written in any instance method.
+
+By default, instance variables are not accessible outside the instance they're attached to. By convention, to make them readable we define a method with the same name as the variable, which returns the current value of that variable. This method is called a _reader_, _getter_ or _accessor_, depending on what language you're in (Ruby uses _reader_).
```
kariabancroft: I wonder if it would be good to elaborate on this "not accessible outside" thing - or if it's ok if they're not quite bought into it yet. Not sure...

01-ruby-fundamentals/07-oo-classes.md
droberts-ada: Maybe we could switch to using the `Date` or `Time` class instead. That would jive with an example we use later too.

01-ruby-fundamentals/07-oo-classes.md
droberts-ada: I'm worried about conflating "helper methods" and "instance methods", since they've got the same word, and because helper methods aren't really visible as methods.

01-ruby-fundamentals/07-oo-classes.md
droberts-ada: It's hard to come up with a good example without a constructor, and we don't introduce those until later (that discussion relies on instance variables). We can probably just let it slide for now.

kariabancroft: Ok - fair

PR #263, opened 2017-07-21
Add an extended example with diagram to the methods lecture

droberts-ada:
## Description
See title! [Original resource for the diagram is here](https://docs.google.com/presentation/d/1KQ7dyY4EMkGDYQNuYjrtoSofKiKaQttEx13R1vG_JyM/edit#slide=id.p)

## Todos
N/A

## Feedback Requested
- Is this too long / complex of an example?
- This example uses `puts` a lot and doesn't `return` at all, which might be the opposite of what we're trying to get them to do (even though it makes sense for this problem).
- Anything that should be added to the diagram? Is it clear what's going on?


01-ruby-fundamentals/methods.md
```diff
@@ -97,6 +97,107 @@ puts result
...
+  print_heart
+
+  puts "my sun and my moon!"
+  print_heart
+
+  puts "Love, #{sender_name}"
+end
+```
+
+You'll have to restart `irb` (again with the `-r` flag) in order for it to pick up the changes, and then we can test `print_love_note` using the same procedure. Note that we get three hearts despite only having typed the heart's `puts` statements once in our code - this is part of the power of methods.
```
kariabancroft: I don't think we can skip over the description of how to pass in the parameters here and how they substitute in the `print_love_note` method. We should have at least one example of invoking the method.

01-ruby-fundamentals/methods.md
```diff
@@ -97,6 +97,107 @@ puts result
...
+
+Next we'll make use of this method... by writing another method! The new method, `print_love_note`, will take two parameters, `sender_name` and `receiver_name`, and it will make use of our `print_heart` method.
+
+**Question:** What will the _signature_ of this method look like? What might it _return_?
+
+Add the following code to `love_note.rb`, below our previous `print_heart` method:
+
+```ruby
+# love_note.rb
+def print_love_note(sender_name, receiver_name)
```
kariabancroft: The example is cute though honestly is feels a little bit yucky to me. Not a big fan of the love note context. Could we make it something with smily face ascii art instead?

kariabancroft:
@droberts-ada Updates look good! Thanks!

PR #264, opened 2017-07-21
Three New Exercises

kariabancroft:
## Description
These changes are part of the work that Jamie and I are doing on creating some new exercises for our introductory curriculum that will be supplemental materials for folks who may need more practice beyond our "regular" assigned materials. Contained here are three new exercises:
1. Conditional practice with small word problems
2. A three wave problem with many different combined conditionals
3. A "fill in the blank" exercise for basic practice tracing/utilizing classes & methods

## Todos
I will update the teachers edition textbook with some potential solutions once these problems are solidified.

Not part of this PR - but there are more exercises that I want to create for methods practice.

## Feedback Requested
Are these something that we'd solely want to use as supplemental support, or could/should they be used within our daily/core curriculum? Is there enough information provided to allow us to hand this off to a student to work on independently?


01-ruby-fundamentals/exercises/admissions-program.md
```diff
@@ -0,0 +1,45 @@
+# Admissions Program
+
+![waves](../images/waves.jpg)
+
+This exercise will be good for you if you want more practice with:
+- creating logical boolean statements
+- practicing Ruby conditional syntax
+- constructing code from a word problem
+- practice debugging complex scenarios
```
droberts-ada: These bullets should probably not start with "practice", since that's how the sentence above leads into them.

01-ruby-fundamentals/exercises/admissions-program.md
```diff
@@ -0,0 +1,45 @@
...
+
+Wild Waves Theme Park needs a program for its admission booths. When visitors to the park come up to the booth to
+purchase their tickets, a worker uses this program to figure out how much to charge them. You will write
+this program.
+
+## Wave 1
+There is one ticket price of $30.00. Senior citizens (age ≥ 65) are given a 50% discount. Write this program as follows.  Declare the needed variables (the person’s age, the base price of a ticket ($30) and the price you will charge). Input the user’s age, compute the price of the ticket and output the result in a formatted way (that is, using a $).  Ensure that the age entered is **valid** (whatever that means to you).
+
+Run your program a few times, asking the user for different ages such as 10, 50, 65, 80 and 0.
```
droberts-ada: I love the emphasis on testing, even if we don't call it that explicitly.

01-ruby-fundamentals/exercises/conditionals.md
```diff
@@ -0,0 +1,22 @@
...
+- practicing Ruby conditional syntax
+- constructing code from a word problem
+
+
+1. As a customer of a store, you are interested in the average cost of items that you've purchase from the store. The data you gather from the user is the total **number** of items you've purchased and the total **amount** of money you've spent so far. Be sure that your code will work even if the total number of items is zero!
+
+1. Write a password guessing program to keep track of how many times the user has entered the password wrong. If it is more than 3 times, print "You have been denied access." and terminate the program. If the password is correct, print "You have successfully logged in." and terminate the program.
+
+1. For some reason, your boss has given you an _odd_ arbitrary math problem to solve. First, you want to get a number from the user. Next, do some math! If the number is odd, change its value so that it is now 3 times it's original value plus 1, otherwise change its value so that it is now half of its original value (rounded _down_).
+
```
droberts-ada: If the number is even and they cut it in half, they won't need to round.

01-ruby-fundamentals/exercises/conditionals.md
```diff
@@ -0,0 +1,22 @@
...
+1. As a customer of a craft store, write some code to determine the cost of yarn based on the length someone desires. The user should provide the distance that they desire (in cm), and your code should return the cost of the yarn based on the distance. Use the following chart to determine the price:
+
+| Distance     | Cost    |
+| :------------- | :------------- |
+| 0 through 100       | 5.00       |
```
droberts-ada: What are the units on the cost? $/cm? Wouldn't the price per unit go down as the quantity increases?

01-ruby-fundamentals/exercises/problem-practice/problem.md
```diff
@@ -0,0 +1,18 @@
+# Math Problem Analyzer
```
droberts-ada: I really like this one! Aside from being good practice, having them work with our starter code is demonstrating good style.

01-ruby-fundamentals/exercises/conditionals.md
kariabancroft: I don't think the expectation is that this is cost per unit. If that were the case, what you're saying makes sense. I think it is a straight cost based on the distance, though I guess that doesn't really model the real world problem that well. (why would you only get 25 if you could get 100 for the same price). I can adjust the problem to clarify and create a cost per unit instead.

kariabancroft:
@droberts-ada I just pushed an update to the yarn/unit question. Let me know if that clarifies.

01-ruby-fundamentals/exercises/problem-practice/problem.md
```diff
@@ -0,0 +1,18 @@
...
+This exercise will be good for you if you want more practice with:
+- writing methods in a class
+- calling methods on an instance of a class
+- working within (some) code that is already written for you
+
+In this exercise, we will use object-oriented principles to create a math problem analyzer. By providing both the text of a problem as well as the numeric answer, your program should be able to prompt your user with the question and provided answer and tell the user whether or not the answer is correct.
+
+## The Code
+There are two files that accompany this exercise. The `runner.rb` file contains the code that you should execute to determine whether or not your class code will work. The `problem.rb` file contains the starter code for the `Problem` class which you will modify. You should not need to modify the code in the `runner.rb` file to get the program to work.
+
```
CheezItMan: Good intro to the concept of testing here too.

PR #265, opened 2017-07-27
Add comparison of classes vs hashes

droberts-ada:
## Description
I've torn out the section on initializing an object from a class, since that's not really something anyone actually does, and replaced it with a discussion of the differences between hashes and objects in Ruby and when it's appropriate to use one vs the other.

## Todos
N/A

## Feedback Requested
Do the guidelines in the last table feel appropriate? Anything that needs to be added?


01-ruby-fundamentals/hashes-classes-and-you.md
```diff
@@ -1,107 +1,106 @@
...
 
-Using a hash to provide parameters to an `initialize` method lets us omit parameters that may not exist for some instances. It also adds clarity to `class` instantiation by explicitely stating keys and values.
+Choosing how your data will be stored is one of the first steps in writing a complex program. Here are some guidelines:
 
-```ruby
-# not using a hash to initialize
-Address.new("Jeremy","","","","WA","","")
+If your data... | Then you should use a...
+---             | ---
+Comes from an external source, like a file or the internet | Hash
```
kariabancroft: Hmmm I'm not sure I agree with this. Maybe we should make these guidelines more flexible. "Then you _could_ use a"... Also, potentially adding a caveat that these thoughts can be a starting point and as they learn more about hashes and classes that these guidelines might evolve.

01-ruby-fundamentals/hashes-classes-and-you.md
```diff
@@ -1,107 +1,106 @@
...
+Involves giving names to pieces of data, without any specific behavior | Hash
+Has complex behaviors you need to model | Class
+Involves many pieces of data following a repeated pattern | Class
 
-# now with a fancy hash!
-Address.new(first_name: "Jeremy", state: "WA")
-```
+<!-- DPR: don't really want to emphasize how confusing things are
```
kariabancroft: Without talking about using hashes to instantiate classes, are we also missing an opportunity to highlight the way that hashes can be passed as "regular" method params?

droberts-ada: Highlighting that is probably a good idea, but I'm not sure it fits into this resource with the way it's been restructured. It might be a good thing to add to the methods resource.

PR #274, opened 2017-08-08
Discuss keyword arguments instead of default values for params

droberts-ada:
## Description
Almost a complete rewrite of the default parameters lecture, to use keyword arguments instead. This is the modern ruby way of doing things, and matches up with the syntax they'll see in Rails.

## Todos
- [ ] Remove hash initialization from Solar System

## Feedback Requested
Is the code sample too complex? I tried to keep it as simple as I could.

### DIFF
--- a/01-ruby-fundamentals/keword-arguments.md
+++ b/01-ruby-fundamentals/keword-arguments.md
@@ -148,7 +148,7 @@ coffee_price(:cappuccino, :grande)
 ```
 
 There are a few things to note here:
-- In the method signature, each keyword argument comes with a _defualt value_
+- In the method signature, each keyword argument comes with a _default value_
   - The syntax is similar to keys and values in a hash
 - When you invoke a method, keyword arguments are optional
   - If you omit one, the value from the method signature will be used instead
--- a/01-ruby-fundamentals/keword-arguments.md
+++ b/01-ruby-fundamentals/keword-arguments.md
@@ -117,11 +117,11 @@ The code is starting to get pretty complicated, but that makes sense because our
 Our _method signature_ is also starting to get complex. Here are two examples of calling the new `coffee_price`:
 
 ```ruby
-# Price a venti iced soy latte with 3 extra shots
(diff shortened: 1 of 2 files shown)


01-ruby-fundamentals/keword-arguments.md
```diff
@@ -0,0 +1,175 @@
...
+
+All of these put the onus on the caller to do things right, which is just asking for trouble. **A well-written method is easy to use right and hard to use wrong** - by that standard, this is not good code.
+
+### Keyword Arguments
+
+One way to address these problems is with a technique called _keyword arguments_. Here's what they look like:
+
+```ruby
+# Define the method
+def coffee_price(type, size, extra_shots: 0, cold: false)
```
kariabancroft: I think it would be good to have one final gist that contains all of the code as it ends up so that there isn't any confusion about the code carryover from above

01-ruby-fundamentals/keword-arguments.md
```diff
@@ -0,0 +1,175 @@
...
+
+# Price for a tall iced drip
+coffee_price(:drip, :tall, cold: true)
+
+# Price for a grande cappuccino
+coffee_price(:cappuccino, :grande)
+```
+
+There are a few things to note here:
+- In the method signature, each keyword argument comes with a _defualt value_
```
CheezItMan: _defualt value_  should be a _default value_

01-ruby-fundamentals/keword-arguments.md
```diff
@@ -0,0 +1,175 @@
...
+
+  # Modify for size
+  if size == :tall
+    # no change
+  elsif size == :grande
+    price *= 1.3
+  elsif size == :venti
+    price *= 1.6
+  else
+    puts "Invalid size: #{size}"
```
droberts-ada: yes

droberts-ada:
https://github.com/AdaGold/Solar-System/pull/8

PR #277, opened 2017-08-09
Remove bignum example

PilgrimMemoirs:
## Description
Removed bignum example - should never use it since ruby's integer class handles it. Unnecessary tangent to have it in this lesson.

## Todos
- [ x] Section X still needs some more detailed examples
- [ x] Still need to add "Additional Resources"

## Feedback Requested
none


droberts-ada:
I think this syntax (`123_456_789`) is still useful, it's just not a Bignum anymore.

PR #280, opened 2017-08-10
Change referral to specific line in code example

vertige:
## Description
A few sentences describing the overall purpose of these textbook changes.

## Todos
- [ ] Section X still needs some more detailed examples
- [ ] Still need to add "Additional Resources"

## Feedback Requested

### DIFF
--- a/01-ruby-fundamentals/methods-worksheet.md
+++ b/01-ruby-fundamentals/methods-worksheet.md
@@ -112,7 +112,7 @@ print_note(user_name, friend_name)
 puts "Thank you for using the friendship note app"
 ```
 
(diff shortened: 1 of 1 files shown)


susanev:
Could we add line numbers and then reference it exactly? We may want to also consider not using words like "above" or "below" as they make the curriculum a bit harder to read for anyone who is using a screen reader.

kariabancroft:
@susanev I really like that idea - unfortunately GItHub's markdown format doesn't provide a way to add line numbers to code samples. I agree the text should be more clear - I think it should reference the code itself rather than the relative line.

susanev:
@kariabancroft yah, lame. I think restating the line itself in the sentence is a good work around.

PR #285, opened 2017-08-14
add git process diagram

kariabancroft:
## Description
A git process diagram for introductory git usage in class. Specifically does not include pull or collaboration information as that will be added in a later diagram.

## Todos
None

## Feedback Requested
Does this encompass what we want @PilgrimMemoirs? Is this good enough for us to try to print poster versions?


kariabancroft:
[Git process diagram to hang in the classrooms](https://trello.com/c/eCsQshVo/352-git-process-diagram-to-hang-in-the-classrooms)

PR #291, opened 2017-08-21
Rework example code in the class methods lecture

droberts-ada:
## Description

Used to be a hodge-podge of different examples, including the Penguin,
students had identified as frustratingly unrealistic.

Now it's all focused on a Song class that represents a song in a music
player program. This should be a little more concrete, and the common
thread should make it easier to follow.

The exception is the last example, where we demonstrate a factory method.
I honestly just couldn't think of a good example for songs that didn't involve
doing a bunch of weird filesystem stuff (i.e. populate music player from the
contents of a directory).

## Todos

N/A

## Feedback Requested

Nothing in particular. I've added Jamie because she volunteered, and Chris because it's his week and someone from Carets ought to have a voice on this.

### DIFF
--- a/02-intermediate-ruby/class-methods-and-self.md
+++ b/02-intermediate-ruby/class-methods-and-self.md
@@ -175,20 +175,26 @@ end
 puts Song.total_plays
 # => 0
 
-s1 = Song.new("Respect", "Aretha Franklin", "songs/respect.mp3")
-s2 = Song.new("What a Little Moonlight Can Do", "Billie Holiday", "songs/moonlight.mp3")
+respect = Song.new("Respect", "Aretha Franklin", "songs/respect.mp3")
+moonlight = Song.new("What a Little Moonlight Can Do", "Billie Holiday", "songs/moonlight.mp3")
 
 3.times do
-  s1.play
+  respect.play
 end
 
 5.times do
-  s2.play
+  moonlight.play
 end
 
-puts "song 1: #{s1.play_count} plays, song 2: #{s2.play_count} plays, total: #{Song.total_plays} plays"
-# => song 1: 3 plays, song 2: 5 plays, total: 8 plays
+puts "#{respect.name}: #{respect.play_count} plays"
+puts "#{moonlight.name}: #{moonlight.play_count} plays"
+puts "total: #{Song.total_plays} plays"
+# Respect: 3 plays
+# What a Little Moonlight Can Do: 5 plays
+# total: 8 plays
 ```
+
+All that `total_plays` does is return the value of `@@total_plays`. If `@@total_plays` were an instance method, we would use the `attr_reader` helper method to accomplish the same thing. Unfortunately there's no equivalent to `attr_reader` for class variables, so we have to do the work ourselves.
 
 #### Questions
 - Consider the class variable `@@total_plays`
--- a/02-intermediate-ruby/class-methods-and-self.md
+++ b/02-intermediate-ruby/class-methods-and-self.md
@@ -10,93 +10,106 @@
 - Examine why we'd use one type of method over another
 
 ## Ruby Classes Thus Far
-### Constructors (`initialize`) & Instance Variables
+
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/class-methods-and-self.md
```diff
@@ -10,171 +10,209 @@
...
 ```
 
 ```ruby
-order = Order.new(1000, 2)
-order.to_money #=> $20.00
+song = Song.new("Respect", "Aretha Franklin", "songs/respect.mp3")
+puts song.summary
+# => Respect, by Aretha Franklin
+song.play
+# 🎶 R-E-S-P-E-C-T 🎶
```
kariabancroft: lol - i love this

kariabancroft: lol - i love this

02-intermediate-ruby/class-methods-and-self.md
```diff
@@ -10,171 +10,209 @@
...
-p = Penguin.new("Gerri")
-puts p.name
-puts p.looks
-```
+# No instances required!
+puts Song.total_plays
+# => 0
 
-Which piece of this class will always be the same for **every instance** of the class?
+s1 = Song.new("Respect", "Aretha Franklin", "songs/respect.mp3")
```
kariabancroft: Could you make these more contextual variable names - since I'd ask the students to do the same

02-intermediate-ruby/class-methods-and-self.md
```diff
@@ -10,171 +10,209 @@
...
-    return "Fuzzy!"
+  def play
+    @play_count += 1
+    @@total_plays += 1
+    # ... load the song data from the file and send it to the speakers ...
+  end
+
+  # Define a class method to access the class variable
+  def self.total_plays
+    return @@total_plays
```
kariabancroft: Right now it seems like this is just accessing a variable, so it might be good to wrap it in a string similar to the summary method so it is doing more than just returning a variable?

02-intermediate-ruby/class-methods-and-self.md
```diff
@@ -10,171 +10,209 @@
...
+- Consider the class variable `@@total_plays`
+  - Where is `@@total_plays` initialized?
+  - How would our program change if we did not initialize this variable?
+  - How would our program change if we initialized this variable in the `initialize` method?
+- Why is `total_plays` a class method? How would our program change if it was an instance method?
 
-  def initialize(name)
-    @name = name
-  end
+### Activity: `Song.most_played`
```
kariabancroft: Nice! I really like this activity

02-intermediate-ruby/class-methods-and-self.md
```diff
@@ -184,26 +222,27 @@ class Pawn
   # This is the class method, it starts with self.
   # It is only called on the class directly Pawn.make_row
   def self.make_row(side)
-    if side == "white"
+    if side == :white
```
kariabancroft: Nice - I like this conversion to symbols

02-intermediate-ruby/class-methods-and-self.md
```diff
@@ -10,171 +10,209 @@
...
-puts test.rad
-# will output 'This is an INSTANCE METHOD'
+instance = ClassName.new
+puts instance.instance_method
+# 'This is an INSTANCE METHOD'
 ```
 
-### Adding in Context
+We've already seen one class method: `new`. `new` is a built-in Ruby method defined on all objects, which builds an instance of that class, calls its `initialize` method, and returns it.
+
```
CheezItMan: I like the connection with `new`, tying it to what they already know.

02-intermediate-ruby/class-methods-and-self.md
droberts-ada: This method doesn't do much, but Ruby doesn't give you the helper method shorthand for class variables, so we definitely need it. I originally meant to spell that out explicitly in the text but I guess it got lost.

PR #292, opened 2017-08-21
Updates to the Modules as Namespaces resource

droberts-ada:
## Description
- Added a substantial introduction, which goes over syntax in detail (addressing a complaint we had from students).
- Use a more "computery" example to introduce namespacing, something that might plausibly come up in a real project.
- Remove the final example of nested modules - this isn't something they're likely to see, and it's an extremely confusing bit of code (even to me!).

## Todos
N/A

## Feedback Requested
- Too heavy of an intro?
- Can you think of a way to demonstrate deeply nested constant lookups realistically without writing a 10000 line library?
- Requested reviews from Jamie because she volunteered, and Chris because he initially authored the library and someone from Carets ought to have a voice on this.

### DIFF
--- a/02-intermediate-ruby/modules-as-namespaces.md
+++ b/02-intermediate-ruby/modules-as-namespaces.md
@@ -1,5 +1,66 @@
 # Modules as Namespaces
 
+## Learning Goals
+By the end of this lesson, students should be able to
+
+- Read and write code using Ruby _modules_
+- Use modules to group classes and methods into _namespaces_
+
+## Modules
+
+Modules are a way to group together related ideas. They're like classes that can't be instantiated, making them a good choice if you need to tie together some methods that don't have any associated state.
+
+We've seen a module before: `Math`. It doesn't makes sense to have an instance of `Math`, it's just a collection of related methods and constants.
+
+```ruby
+# No instances required!
+Math::PI
+# => 3.141592653589793
+Math.sqrt(25)
+# => 5.0
+Math.cos(Math::PI / 4)
+# => 0.7071067811865476
+```
+
+### Syntax
+
+#### Defining a Module
+
+Defining a module is very similar to defining a class, except they use Ruby's `module` keyword.
+
+```ruby
+module ModuleName
+  # Define a constant
+  CONSTANT_NAME = 4
+
+  # Define a method
+  def method_name
+    # ...
+  end
+
+  # Define a class
+  class ClassName
+    def initialize(a, b)
+      # ...
+    end
+
+    def instance_method
+      # ...
+    end
+  end
+end
+```
+
+#### Using a Module
+
+Modules can contain constants, methods, classes and even other modules. The syntax for accessing something is a module is very similar to accessing something in a class.
+- Constants, classes and other modules are accessed with `::`
+    - `ModuleName::ClassName`
+    - `Math::PI`
+- Methods are accessed with `.`
+    - `ModuleName.method_name`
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/modules-as-namespaces.md
```diff
@@ -1,48 +1,115 @@
...
+- Read and write code using Ruby _modules_
+- Use modules to group classes and methods into _namespaces_
 
-For example we might create a class called "Shape" in a graphics program, but that name might be used elsewhere in another library.  When you do Shape.new, how is Ruby to know which Shape class to instantiate?  It's a similar idea to putting files into directories.  Files in the same directory cannot have the same name, but files in different directories may.  
+## Modules
 
-In this example below we can create an Array class and put it into a module.  Notice that it does not affect Ruby's default Array class at all
+Modules are a way to group together related ideas. They're like classes that can't be instantiated, making them a good choice if you need to tie together some methods that don't have any associated state.
+
+We've seen a module before: `Math`. It doesn't makes sense to have an instance of `Math`, it's just a collection of related methods and constants.
```
kariabancroft: I might re-word this section to exclude the use of "just": The designers of Ruby decided to make `Math` a module rather than a class because it is a collection of...

02-intermediate-ruby/modules-as-namespaces.md
```diff
@@ -1,48 +1,115 @@
...
+# => 5.0
+Math.cos(Math::PI / 4)
+# => 0.7071067811865476
+```
+
+### Syntax
+
+#### Defining a Module
+
+Defining a module is very similar to defining a class, except they use Ruby's `module` keyword.
```
kariabancroft: Might be good to add that any/all of these pieces of syntax are optional and not all required within a module.

02-intermediate-ruby/modules-as-namespaces.md
```diff
@@ -1,48 +1,115 @@
...
+```
 
-our_array = Perimeter::Array.new
-ruby_array = Array.new
+#### Using a Module
 
-p our_array.class
-p ruby_array.class
-```
+Modules can contain constants, methods, classes and even other modules. The syntax for accessing something is a module is very similar to accessing something in a class.
```
kariabancroft: Per my comment above, I think it'd move this first sentence to above and keep the second one here with the examples.

02-intermediate-ruby/modules-as-namespaces.md
```diff
@@ -1,48 +1,115 @@
...
-The two "Array" classes can work alongside each other.  This is because we namespaced our Array class under the Perimeter Module.  
+## Namespacing
 
-The :: operator looks up objects and classes within a module, in this case we use it to look up the Perimeter Module's Array class.  
+*Namespacing* is a way of bundling related objects together. In ruby we use Modules to group related Classes and methods together. We do this to avoid naming collisions.
 
-Leaving off the module makes us extend the Array class globally which could lead to side-effects if not every part of our program does not expect the new behavior.  
+For example, imagine we're building an inventory management system for an arts and crafts store. We might want a class called `String`, but that name is already taken by one of Ruby's built-in classes.  When you do `String.new`, how is Ruby to know which `String` class to instantiate? The solution is to put our class in a module.
+
+It's a similar idea to putting files into directories.  Files in the same directory cannot have the same name, but files in different directories may.
```
kariabancroft: Yes! Great!

PR #293, opened 2017-08-23
very basic while loop examples worksheet

kariabancroft:
## Description
Per Dan's note about not having any worksheet practice with while loops. This will go after "complex conditionals" lecture.

## Todos
None

## Feedback Requested
I put this in a separate file, since the conditionals practice was solely about evaluating many conditional expressions together.

Are these too simple?

### DIFF
--- a/01-ruby-fundamentals/exercises/while-worksheet.md
+++ b/01-ruby-fundamentals/exercises/while-worksheet.md
@@ -2,10 +2,9 @@
 For this exercise, you will be evaluating some Ruby code and coming up with scenarios for how the code will execute.
 
 1. For the code below, describe what the loop will do with the following `word` variable assignment. Be sure to include how many times the loop will execute.
-
-  a. `word = "ham"`  
-  b. `word = "hamster"`  
-  c. `word = "bananas"`  
+  - `word = "ham"`  
+  - `word = "hamster"`  
+  - `word = "bananas"`  
 
   ```ruby
   while word.length > 7
@@ -15,10 +14,9 @@ For this exercise, you will be evaluating some Ruby code and coming up with scen
   ```
 
 1. For the code below, how many times will the block of the loop execute with the following `word` variable assignments? What is the final value for the `i` variable for each?
-
-  a. `word = "ham"`  
-  b. `word = "hamster"`  
-  c. `word = "harmonious"`  
+  - `word = "ham"`  
+  - `word = "hamster"`  
+  - `word = "harmonious"`  
 
   ```ruby
   i = 0
--- a/01-ruby-fundamentals/exercises/while-worksheet.md
+++ b/01-ruby-fundamentals/exercises/while-worksheet.md
@@ -11,7 +11,7 @@ For this exercise, you will be evaluating some Ruby code and coming up with scen
   puts "Note: this word must be less than or equal to 7 characters."
 
   word = gets.chomp
-  while word > 7
+  while word.length > 7
     puts "Please enter a word less than or equal to 7 characters."
     word = gets.chomp
   end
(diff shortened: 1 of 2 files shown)


kariabancroft:
[C9: Week 1 conditionals worksheet needs section on while loops](https://trello.com/c/no4gdjSi/362-c9-week-1-conditionals-worksheet-needs-section-on-while-loops)

01-ruby-fundamentals/exercises/while-worksheet.md
```diff
@@ -0,0 +1,55 @@
+# While Loop Evaluation Worksheet
+For this exercise, you will be evaluating some Ruby code and coming up with scenarios for how the code will execute.
+
+1. For the code below, describe what the code will do with the following inputs:
+  1. "ham"
+  2. "hamster"
+  3. "bananas"
```
droberts-ada: This is pretty open-ended. Can you make the prompt more specific?

01-ruby-fundamentals/exercises/while-worksheet.md
```diff
@@ -0,0 +1,55 @@
...
+  word = gets.chomp
+  while word > 7
+    puts "Please enter a word less than or equal to 7 characters."
+    word = gets.chomp
+  end
+  ```
+
+1. For the code below, how many times will the block of the loop execute with the following inputs for the `word` variable:
+  1. "ham"
+  2. "hamster"
```
droberts-ada: Maybe also ask about what the final value of `i` will be.

01-ruby-fundamentals/exercises/while-worksheet.md
```diff
@@ -0,0 +1,55 @@
...
+  i = 0
+  while i < word.length
+    puts word[i]
+    i += 3
+  end
+  ```
+
+1. For the code below, describe the condition where the loop will terminate.
+  ```ruby
+  time = 0
```
droberts-ada: Won't this never terminate? Is that the intention?

kariabancroft: Correct - that is the intention.

kariabancroft:
@droberts-ada I updated per your comments

01-ruby-fundamentals/exercises/while-worksheet.md
```diff
@@ -0,0 +1,55 @@
+# While Loop Evaluation Worksheet
+For this exercise, you will be evaluating some Ruby code and coming up with scenarios for how the code will execute.
+
+1. For the code below, describe what the loop will do with the following input to the `word` variable. Be sure to include how many times the loop will execute.
+  1. "ham"
+  2. "hamster"
+  3. "bananas"
```
droberts-ada: Looking at this again, I guess I'm still confused about this one. Are these the words that are provided by the user for one run of the program, or do these represent 3 separate runs? In either case I'm not sure they make sense:
- If it's all the words for one run, then only the first will be used
- If it's 3 separate runs, then what happens in the last run after "bananas" is rejected?

kariabancroft: They represent 3 separate runs. For the last one, I want them to identify that it will prompt for a new word. I don't want to put multiple words in because I think it gives away the result.

droberts-ada: What if you gave a sequence of words for each run, and called out explicitly that not all words will necessarily be used?

kariabancroft: Hmmmm - I'm not really convinced that is less confusing than what I have now. But if you think it is I'll do that.

PR #298, opened 2017-08-24
Variety of updates on Week 4 curriculum

kariabancroft:
## Description
Some general cleanup when reviewing the week 4 curriculum.

## Todos
N/A

## Feedback Requested
Nothing specific

### DIFF
--- a/02-intermediate-ruby/enumerable.md
+++ b/02-intermediate-ruby/enumerable.md
@@ -113,10 +113,10 @@ Use these methods to filter collections by a conditional evaluation. `select` wi
 ```
 
 ## There's so much more to discover
-`Enumerable` is one of the most exciting things about Ruby. It is immensely useful. With it, we can write expressive, highly semantic code that is also concise. Spend some time with the [Enumerable docs](http://ruby-doc.org/core-2.4.0/Enumerable.html). Good knowledge in here will serve you in every application you build, and not just with Ruby. The methods contained in `Enumerable` describe ideas and patterns you'll find in every programming discipline.
+`Enumerable` is one of the most exciting things about Ruby. It is immensely useful. With it, we can write expressive, highly semantic code that is also concise. Spend some time with the [Enumerable docs](http://ruby-doc.org/core-2.4.1/Enumerable.html). Good knowledge in here will serve you in every application you build, and not just with Ruby. The methods contained in `Enumerable` describe ideas and patterns you'll find in every programming discipline.
 
 ## Activity: Refactor Loops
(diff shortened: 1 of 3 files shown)


kariabancroft:
[Random Week 4 curriculum updates](https://trello.com/c/8ydHhSkq/384-random-week-4-curriculum-updates)

kariabancroft:
@tildeee We definitely can! I think it's be good to highlight how it is similar/different than the type of object composition we have done thus far. We don't really use the `include` syntax beyond this small example, so I'm fine sort of breezing over it. We could add an "additional resources" link though if you think that might be useful.

02-intermediate-ruby/enumerable.md
```diff
@@ -113,10 +116,10 @@ Use these methods to filter collections by a conditional evaluation. `select` wi
 ```
 
 ## There's so much more to discover
-`Enumerable` is one of the most exciting things about Ruby. It is immensely useful. With it, we can write expressive, highly semantic code that is also concise. Spend some time with the [Enumerable docs](http://ruby-doc.org/core-2.4.0/Enumerable.html). Good knowledge in here will serve you in every application you build, and not just with Ruby. The methods contained in `Enumerable` describe ideas and patterns you'll find in every programming discipline.
+`Enumerable` is one of the most exciting things about Ruby. It is immensely useful. With it, we can write expressive, highly semantic code that is also concise. Spend some time with the [Enumerable docs](http://ruby-doc.org/core-2.4.1/Enumerable.html). Good knowledge in here will serve you in every application you build, and not just with Ruby. The methods contained in `Enumerable` describe ideas and patterns you'll find in every programming discipline.
```
droberts-ada: If you change this URL to `http://ruby-doc.org/core/Enumerable.html` (leaving out the `-2.4.1`), it will redirect to the most recent version, and we won't have to update this document every cohort.

02-intermediate-ruby/more-enumerable.md
```diff
@@ -38,13 +38,12 @@ scores = {
...
 ```
 
-Use Enumerable methods to answer the three questions posed in the comments. After we've had some time to explore, we'll review solutions as a class.
+Use Enumerable methods to answer the questions below. After we've had some time to explore, we'll review solutions as a class.
+
+**Questions**
+1. Who has the single highest score (and what is that score)?
+1. Who has the highest *average* score (and what is that average)?
+1. Average is the sum of scores divided by the number of scores
+1. Who has the biggest difference between their highest and lowest score (and what is that difference)?
```
droberts-ada: I think that line 48 should be a sub-bullet, something like

```md
1. Who has the single highest score (and what is that score)?
1. Who has the highest *average* score (and what is that average)?
    - Average is the sum of scores divided by the number of scores
1. Who has the biggest difference between their highest and lowest score (and what is that difference)?
```

droberts-ada:
@tildeee I think I agree that spending a bunch of time on mixins is not very valuable, since they don't come up much in the context of Rails. Especially if we're coming back to inheritance when we eventually get to Rails, it should be OK to leave out the demonstration here.

PR #300, opened 2017-08-25
Pass at changing Enumerable example from String Smiley to Messageable

tildeee:
## Description
I feel that the current example of Mixins is too abstract to be meaningful for students.

I wanted to provide a different example, particularly one that uses the "-able" naming convention (like "Enumerable" or "Comparable")

In general, I also am looking for ways to de-emphasize comprehending "Mixins," as the real core of this lesson is Enumerables anyway.

## Feedback Requested
- Definitely let me know if the example is still not perfect (yes, I made up the word "Messageable")
- I'm not sure if I'm adding too much information referring back to design/object composition
- Still trying to de-emphasize mixins/emphasize Enumerable and its functions


02-intermediate-ruby/enumerable.md
```diff
@@ -19,35 +19,50 @@ class String # <= wat?!? We can do this? We sure can!
...
+
+  def initialize
+    @inbox_count = 0
+  end
 end
 ```
 
-__Question: Could we `include AllSmiles` in class `Integer`? Why?__
+Here, we use mixin functionality by using `include` on a module (Messageable) within a class. When we `include` a module, all of the module's instance methods become available to that class.
+With this design decision, we are choosing to design classes so that they have an _acts-like_ relationship. An SMSInbox _acts-like_ it is messageable, as does an EmailInbox. Using an _acts-like_ relationship doesn't lock us down into a hierarchical relationship that inheritance/is-a relationship might.
```
kariabancroft: Yes - this is great

02-intermediate-ruby/enumerable.md
```diff
@@ -126,3 +141,6 @@ In a group of 4, you shall research each of these `Enumerable` methods. Prepare
 - `group_by`
 - `zip`
 - `drop_while` & `take_while`
+
+## Additional Resources
+- For another interesting Mixin, look up [Comparable](https://ruby-doc.org/core-2.4.0/Comparable.html)
```
kariabancroft: Dan had a recommendation for me recently to remove the version number and just put "/core/" which will automatically point to the latest version

tildeee:
pinging @droberts-ada  :) 🌮

PR #309, opened 2017-09-20
update on params

tildeee:
Shouldn't the params just be for [:title] and [:author] instead of [:book][:author]?


tildeee:
lol nevermind I see what I'm doing wrong

PR #311, opened 2017-09-25
AR Relationships, move file, add resources exercise

kariabancroft:
## Description
- Switch the context of the AR relationships notes from artists/albums to authors/books. 
- Add a new resource which goes through the details of updating the existing books app with new relationships.
- Move nested resource to regular rails folder from intermediate.
- Add an exercise to the route resources notes
- Clean up agile notes

## Todos
N/A

## Feedback Requested
Should there be another resource for setting up the relationships from scratch in a brand new Rails app? Does it make sense that I split out these instructions into a separate document?


kariabancroft:
[Put AR Relationships Lecture in Book format](https://trello.com/c/vKGxU17S/436-put-ar-relationships-lecture-in-book-format)

08-rails/README.md
```diff
@@ -20,6 +20,7 @@
 1. [DB Seeds](./db-seeds.md)  
 1. [Routes with Resources](./routes-and-resources.md)  
 1. [Rails Debugging](./rails-debugging.md)  
+1. [Nested Routes](./nested-routes.md)
```
tildeee: thanks so much for updating this 🎆 🙇

tildeee:
As for the feedback you requested in this PR (just saw that section): I think it'd be fine if you put it in a different document. The cons are that it means that it'll be another thing in the repo, but at least it'd make sure that the lecture (assuming that these notes are mostly used to structure lecture) stays focused on the activity! Then we can just link to that second document at the bottom of the current lecture notes.

tildeee:
Just realized that this activity requires the instructors to spend time in class to cover the following after it:
- updating the views to display `book.author.name` instead of `book.author`
- updating the structure to create authors and to use authors during book creation

kariabancroft:
@tildeee Per your comments about the views being updated - I think I intentionally skipped over this in class because it truly takes forever. I'll write up these instructions as part of the exercise and then I think we can skip them as part of the in-class time. Does that seem reasonable?

PR #318, opened 2017-10-19
Rails API Updates

kariabancroft:
## Description
These changes encompass the second portion of the intermediate Rails curriculum. Updates include:
- A non-pokemon example
- Some verbiage updates
- More explicit TDD approach in creating API lecture
- Remove cohort-specific links and references

## Todos
None

## Feedback Requested
Does the Dog class with the Dog.ceo example seem OK? The endpoint isn't super interesting but I think that might help just focus the learning goals.


09-intermediate-rails/creating-apis.md
```diff
@@ -62,30 +62,34 @@ You'll notice that we have one basic get route created for pets, and one basic c
...
 def index
-  render :json => { ready_for_lunch: "yassss" }
+  render json: { ready_for_lunch: "yassss" }
 end
 ```
 
-Notice that we didn't for realsies write any JSON. We provided a plain Ruby hash and let Rails do the conversion for us (with the `render :json =>` call. So to make progress on our tests, we could do something like:
+Notice that we didn't for realsies write any JSON. We provided a plain Ruby hash and let Rails do the conversion for us (with the `render json: ` call. When we run `rails test` again we see that our test is expecting an `Array` rather than a Ruby `Hash` which is what our controller method contains.
+
+To create the appropriate data and ensure we send back an array, we could do something like:
```
droberts-ada: 3 things:

1. You have extra whitespace on `render json: `
1. The following sentence ("When we run...") should probably be a new paragraph.
1. I'm not sure I understand the last sentence - maybe reword it, or at least use a different verb than "contains"

09-intermediate-rails/creating-apis.md
```diff
@@ -166,7 +171,19 @@ def pet_params
 end
 ```
 
-As before, the big difference from a web app is in what we send back. While in a web app we would probably redirect the client to the resource they created, that doesn't make sense for an API. Instead, we just send back some JSON containing the ID of the newly created pet.
+As before, the big difference from a web app is in what we send back. While in a web app we would probably redirect the client to the resource they created, that doesn't make sense for an API. Instead, we just send back some JSON containing the ID of the newly created pet as well as the appropriate status code.
+
+**ALERT!** Rails is built with CSRF protection which essentially prevents malicious requests. You might remember seeing notes in passing about this in our forms curriculum, where we had to explicitly add an _authenticity token_ to each form.
+
+Since preventing these cross site requests doesn't apply in an API context, we will add a Rails helper method to our controller to allow these `POST` requests to go through:
```
droberts-ada: Should CSRF be its own section, with a full heading and everything? I can see how it fits here, but I think if I was a student trawling back through textbook pages trying to figure out why my capstone API doesn't work this would be hard to spot.

droberts-ada: Also, have you confirmed this is the only thing you need to do to get around CSRF?

09-intermediate-rails/consuming-apis-in-rails.md
```diff
@@ -7,9 +7,7 @@
 - Use data from an API in a Rails site
 
 ## Initial Setup
-
-To start clone or fork and clone the [initial project setup](https://github.com/Ada-C7/slack_api_example).
-
+In the calendar, you'll find a link to a github repo. Go ahead and clone that repo!
```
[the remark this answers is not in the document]
kariabancroft: I added a Trello card with a note about adding it to the calendar for that week - so once we have it done once we should be able to continue including it

09-intermediate-rails/creating-apis.md
kariabancroft: Sure I can separate it out into its own section - yes I ran through the example as I updated the notes

kariabancroft:
@droberts-ada I pushed updates per your review comments

PR #322, opened 2017-10-25
Consuming APIs Notes

kariabancroft:
## Description
More updates to the Consuming API notes for:
- Ruby naming conventions
- Updates to provide more thorough code examples
- Clear descriptions for when we are doing live code together and when they'll break out to do an exercise

**Note that these updates go along with a new repo https://github.com/AdaGold/SlackAPI**

## Todos
N/A

## Feedback Requested
Specifically for @tildeee: I want to make sure that with the inclusion of a little hands on activity that we have enough time on the schedule for this resource. Probably 1.5 -2 hours to get through all of it is what we'll need.

### DIFF
--- a/09-intermediate-rails/consuming-apis-in-rails.md
+++ b/09-intermediate-rails/consuming-apis-in-rails.md
@@ -153,7 +153,7 @@ def self.send_msg(channel, msg, token = nil)
 end
 ```
 
-Verify it works through the rails console: `SlackApiWrapper.send_msg("@<username>", "test test test")`
+Verify it works through the rails console: `SlackApiWrapper.send_msg("test-api-channel", "test test test")`
 
 ### The Controller
 
--- a/09-intermediate-rails/consuming-apis-in-rails.md
+++ b/09-intermediate-rails/consuming-apis-in-rails.md
@@ -136,11 +136,10 @@ end
 And finally, `send_msg`:
 
 ```ruby
-def self.send_msg(channel, msg, token = nil)
-  token = TOKEN if token == nil
+def self.send_msg(channel, msg)
   puts "Sending message to channel #{channel}: #{msg}"
 
-  url = BASE_URL + "chat.postMessage?" + "token=#{token}"
+  url = BASE_URL + "chat.postMessage?" + "token=#{TOKEN}"
   response = HTTParty.post(url,
   body:  {
     "text" => "#{msg}",
@@ -224,7 +223,7 @@ end
 private
 
 def self.create_channel(api_params)
-    Channel.new(
+    return Channel.new(
       api_params["name"],
       api_params["id"],
       {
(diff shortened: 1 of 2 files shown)


kariabancroft:
[Consuming API Rails Updates](https://trello.com/c/8MgMweSH/482-consuming-api-rails-updates)

09-intermediate-rails/consuming-apis-in-rails.md
```diff
@@ -148,15 +133,15 @@ def self.listchannels()
 end
 ```
 
-And finally, `sendmsg`:
+And finally, `send_msg`:
 
 ```ruby
-def self.sendmsg(channel, msg, token = nil)
+def self.send_msg(channel, msg, token = nil)
   token = TOKEN if token == nil
```
droberts-ada: Why do we have an argument for `token` here? I don't think this is ever used in the controller - we should omit it and use `TOKEN` everywhere.

If we want to keep it, we should say `token=TOKEN`, rather than having the line with the `if` statement.

09-intermediate-rails/consuming-apis-in-rails.md
```diff
@@ -219,24 +209,43 @@ And modifications to our API wrapper:
...
     end
   end
   return channel_list
 end
+
+private
+
+def self.create_channel(api_params)
+    Channel.new(
+      api_params["name"],
```
droberts-ada: Could you add an explicit return here?

09-intermediate-rails/consuming-apis-in-rails.md
```diff
@@ -219,24 +209,43 @@ And modifications to our API wrapper:
...
+    )
+  end
 ```
 
+Now, all of the logic that relies on the specific Slack API channel data is encapsulated into a single method `create_channel`. If the data in the API response changes, this will be the only method we need to update.
+
+When you reload the root route now, what do you notice? Is there anything else we need to update? (Spoiler alert: YES!)
+
+Once you update the index view, you should now see that
+
```
droberts-ada: What should we now see ❓❓❓

droberts-ada:
The new repo looks good too.

09-intermediate-rails/consuming-apis-in-rails.md
kariabancroft: I will remove this - it was one of the leftovers that I was ignoring for no reason other than, I was.

09-intermediate-rails/consuming-apis-in-rails.md
kariabancroft: oops!

kariabancroft:
@droberts-ada I addressed your comments and updated my solutions branches on the repo per these updates as well.

PR #328, opened 2017-11-06
Cm/es6

CheezItMan:
## Description
Just a look at the Ajax notes to add a bit of arrow functions and a note about using them.  I also added a bit on default arguments for functions (new in ES6).

@droberts-ada 

I've added some changes to make the linter happy with AirBNB standards.

## Todos
- Nope

## Feedback Requested
Please any feedback is welcome.


10-JavaScript/fun-with-functions.md
```diff
@@ -42,6 +42,19 @@ sayItTwice("JS is OK!"); // "JS is OK!"
                          // "JS is OK!"
 ```
 
+##### Default Function arguments
+
+Sometimes, you may want to provide a default value in case a user does not specify an argument.  Just with with Ruby, JavaScript functions can take default Arguments.  You specify the default value for a parameter by setting the variable equal to a value in the function header's parentheses.
```
Hamled: "Just with with Ruby" -> "Just as with Ruby". We should probably also change "parameter" in the last sentence to "argument" to keep it consistent with the rest of this section.

10-JavaScript/ajax-intro.md
```diff
@@ -113,11 +113,13 @@ $.get(url, successCallback);
 Oftentimes, you'll see the above collapsed into one function call:
 ```javascript
 $.get('https://petdibs.herokuapp.com/pets',
-  response => {
+  (response) => {
     console.log('success!');
   });
 ```
 
+Notice our use of arrow functions here.  Because of their compact syntax and the fact that an arrow function gets `this` from it's outside container, they often make great callback functions.
```
Hamled: Arguably the more important of the two benefits for arrows functions as callbacks is the way it handles `this`. If possible I think it would be helpful for the examples above to involve some use of `this`.

Hamled: Additionally, the code sample below (in the "Let's add callbacks..." section) should be updated to use the same arrow syntax as we have in the sample above.

kariabancroft: Seconded, maybe access/utilize the `response` obj

10-JavaScript/fun-with-functions.md
```diff
@@ -42,6 +42,19 @@ sayItTwice("JS is OK!"); // "JS is OK!"
                          // "JS is OK!"
 ```
 
+##### Default Function arguments
+
+Sometimes, you may want to provide a default value in case a user does not specify an argument.  Just as with Ruby, JavaScript functions can take default Arguments.  You specify the default value for a parameter by setting the variable equal to a value in the function header's parentheses.
+
+const sayItTwice = function(text = "JS is Awesome") {
```
kariabancroft: Missing the start of the code block

10-JavaScript/fun-with-functions.md
```diff
@@ -42,6 +42,19 @@ sayItTwice("JS is OK!"); // "JS is OK!"
                          // "JS is OK!"
 ```
 
+##### Default Function arguments
```
kariabancroft: I think it'd make this header one size larger since this way is basically just bolded.

10-JavaScript/fun-with-functions.md
```diff
@@ -42,6 +42,19 @@ sayItTwice("JS is OK!"); // "JS is OK!"
...
+
+Sometimes, you may want to provide a default value in case a user does not specify an argument.  Just as with Ruby, JavaScript functions can take default Arguments.  You specify the default value for a parameter by setting the variable equal to a value in the function header's parentheses.
+
+const sayItTwice = function(text = "JS is Awesome") {
+  console.log(text);
+  console.log(text);
+};
+
+sayItTwice(); // "JS is Awesome"
```
kariabancroft: Can you also add the example for what happens when a parameter is provided?

10-JavaScript/fun-with-functions.md
```diff
@@ -133,15 +133,15 @@ data.forEach(function (num) {
...
 ### Functional Expressions vs. Function Declarations
 Something that may come up in your exploration of JavaScript is the difference between a functional expression and a function declaration. A functional expression is what we've seen so far:
 
 ```javascript
 // functional expression - this is the right way
-const foo = function(bar) {};
+const foo = function foo(bar) {};
 ```
 
 A functional declaration looks like this. Note the lack of `let` and a semicolon, and the difference word order.
```
Hamled: "difference" -> "different"

10-JavaScript/welcome-to-js.md
```diff
@@ -152,6 +152,8 @@ console.log("#{e} hotdog"); // "#{e} hotdog" js doesn't do interpolation with ""
 console.log(`${e} hotdog`);  // `elephant hotdog`
 ```
 
+**Note** We will standardize on using single quotes '' for strings, except in cases of string interpolation (where backticks are acceptable).
```
Hamled: "acceptable" -> "required"

10-JavaScript/fun-with-functions.md
```diff
@@ -133,15 +133,15 @@ data.forEach(function (num) {
 });
 
 let average = sum / data.length;
-console.log("Average is " + average);
+console.log(`Average is ${average}`);
 ```
 
 ### Functional Expressions vs. Function Declarations
```
Hamled: In this section we are using the phrase "functional expression" however I believe the correct terminology is just "function expression".

Hamled: Shortly after this section, in the paragraph after the code snippet for the function declaration example, we say to always use `let` when defining functions and we should probably say `const` instead. (I can't comment on that line directly because it was not changed.)

10-JavaScript/ajax-post.md
```diff
@@ -8,14 +8,14 @@ Just like we used AJAX to send GET requests, we can also use it to send POST req
...
-let url = 'https://petdibs.herokuapp.com/pets';
-let data = {
+const url = 'https://petdibs.herokuapp.com/pets';
+const data = {
   name: 'Chestnut',
-  age: 17
+  age: 17,
 };
-let callback = function(){
+let callback = function callback() {
```
Hamled: Should this be `const` instead of `let`?

Hamled:
Looks good, I think using the linter for all of our code snippets (and any live coding and project starter code) going forward will be a big win!

CheezItMan:
Changes updated as per Hamled

10-JavaScript/fun-with-functions.md
```diff
@@ -136,22 +136,22 @@ let average = sum / data.length;
 console.log(`Average is ${average}`);
 ```
 
-### Functional Expressions vs. Function Declarations
+### Function Expressions vs. Function Declarations
 Something that may come up in your exploration of JavaScript is the difference between a functional expression and a function declaration. A functional expression is what we've seen so far:
```
Hamled: We should also change "functional" to "function" throughout this whole section.

10-JavaScript/classes.md
```diff
@@ -44,18 +44,29 @@ When you create a new book instance the `initialize` method runs and sets the in
 Below is this Book class written in JavaScript.  It has a title, and author attributes and a `toString` method which outputs the object as a String.  Notice that everything in a class falls between the curly braces without any commas to separate each item.  
 
 ```javascript
+/**
+  *  Book class
+  */
 class Book {
```
droberts-ada: I don't know if we need to go quite so far as to provide docstrings.
- I think it'll be really easy to forget to do, and then our curriculum will be inconsistent
- For a small demonstration class like this, the weird syntax might actually distract from what we're trying to teach

10-JavaScript/fun-with-functions.md
```diff
@@ -15,12 +15,12 @@ All functions in JavaScript are variables. Just like any other variable, you dec
...
 ```javascript
-const bark = function() {
-  console.log("Woof!");
+const bark = function bark() {
+  console.log('Woof!');
 };
 ```
 
-**Note** we declared the function `bark` here with `const`.  We could have used `let` or `var`, but because we don't want `bark` to be reassigned, it makes sense to make it a constant.
+**Note** we declared the function `bark` here with `const`.  We could have used `let` or `var`, but because we don't want `bark` to be reassigned, it makes sense to make it a constant.  We also added the function name to the right of the `function` keyword.  This is optional, but helps identify things in the stack trace.
```
droberts-ada: I think the stack trace should be fine either way. AFAIA, naming your function variables is only useful if you want to use recursion:

https://stackoverflow.com/questions/34917414/var-name-function-name-function-name-used-twice

10-JavaScript/classes.md
CheezItMan: Yeah, I'd left one example in here intending to explain a bit about docstrings, but it's does distract from the main points.

Removed.

PR #329, opened 2017-11-07
Rails Templates

kariabancroft:
## Description
Move Rails templates from individual gists to textbook curriculum resources.

## Todos
N/A

## Feedback Requested
NOTE: I did not include awesome print which was only included in one of the template files. should it be included overall?


kariabancroft:
[C9: add .rails-template to end of Week 3 Rails or beginning of Week 4 Rails, transfer .rails-template to ada account](https://trello.com/c/zgpsGN6A/442-c9-add-rails-template-to-end-of-week-3-rails-or-beginning-of-week-4-rails-transfer-rails-template-to-ada-account)

kariabancroft:
@droberts-ada @CheezItMan OK I've just added jquery to both (so that its there at first) and then set the foundation gem version in the second template. Not at all thinking about or prioritizing yarn right now to be honest.

CheezItMan:
Meh, we may be using Go instead of Rails by that time anyway!

PR #330, opened 2017-11-08
JS Worksheets

kariabancroft:
## Description
Updated some of the JS worksheets per updated naming conventions and worksheet adjustments.

Also see the [PR on the TET](https://github.com/Ada-Developers-Academy/Teachers-Edition-Textbook/pull/21).

## Todos
N/A

## Feedback Requested
1. Are the solutions appropriately formatted?
1. Are these worksheets assigned in the appropriate order?
1. Do we need additional worksheets for ES6-specifics?

### DIFF
--- a/10-JavaScript/exercises/functions-worksheet.md
+++ b/10-JavaScript/exercises/functions-worksheet.md
@@ -22,11 +22,11 @@ end
 
 3.
 ```ruby
-def larger_num(num1, num2)
-  if num1 >= num2
-    num1
-  elsif num2 > num1
-    num2
+def larger_num(first, second)
+  if first >= second
+    first
+  elsif second > first
+    second
   end
 end
 ```
@@ -43,29 +43,12 @@ def output(items)
 end
 ```
 
-6.  
-```ruby
-class Chair
-  def initialize(name)
-    @name = name
-  end
-
-  def name
-    return @name
-  end  
-
-  def weight_in_lbs
-    return 29
-  end
-end
-```
-
 
 ### Part 2
 1.
 Invoke the zombies function.
 ```javascript
-var zombies = function() {
+const zombies = function() {
   return "We like to eat people";
 };
 ```
@@ -73,12 +56,12 @@ var zombies = function() {
 2.
 Call the `square()` method on the `my_calculator` object.
 ```javascript
-var Calculator = function() {
+const Calculator = function() {
   this.square = function (x) {
     return x * x;
   }
 };
-var my_calculator = new Calculator();
+let my_calculator = new Calculator();
 ```
 
 3.
@@ -92,12 +75,6 @@ console.log(function () {
 4.
 What does the following code print to the console?
 ```javascript
-var blabbermouth = function() { };
+const blabbermouth = function() { };
 console.log(blabbermouth.name);
 ```
-
-5.
-Create a class called `Rectangle` with constructor function to create objects representing rectangles with length and width properties and a method to calculate the area. Use the `Rectangle` class to create an instance and print the area of the rectangle to the console.
-
-6.
-Now create a subclass of `Rectangle` called `Square` which is a rectangle with equal length and width sides and has a constructor which only takes one argument.
--- a/10-JavaScript/exercises/objects-worksheet.md
+++ b/10-JavaScript/exercises/objects-worksheet.md
@@ -1,5 +1,6 @@
 # JavaScript Worksheet: Objects
 
+## Part 1
 1. What does the following code print to the console?
 ```javascript
 let pen = {};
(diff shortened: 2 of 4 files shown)


kariabancroft:
[Teachers edition textbook for JS worksheets](https://trello.com/c/WNgATDKF/512-teachers-edition-textbook-for-js-worksheets)

10-JavaScript/exercises/functions-worksheet.md
```diff
@@ -43,42 +43,25 @@ def output(items)
...
   return "We like to eat people";
 };
 ```
 
 2.
 Call the `square()` method on the `my_calculator` object.
 ```javascript
-var Calculator = function() {
+const Calculator = function() {
   this.square = function (x) {
```
droberts-ada: If we've got an example of prototypical objects here, do we want to add a classical object example as well?

10-JavaScript/exercises/ruby-to-js-worksheet.md
```diff
@@ -45,13 +43,6 @@ Some of the later problems use Ruby `each` loops. For now, solve them using `for
 
 1. &nbsp;
     ```ruby
-    (1..3).each do |i|
-      puts i * i
-    end
```
droberts-ada: What is your reasoning for taking these problems out? Too many questions asking the same thing?

10-JavaScript/welcome-to-js.md
```diff
@@ -243,6 +243,29 @@ To point out the three components:
 
 **Question:** What Ruby code could you use to achieve the same thing?
 
+#### more `for` Loops
+There are a few other types of for loops that are useful. These are a bit more similar to the `each` loops we are used to seeing in Ruby.
+
+**`for`...`in`...**
+
+This `for` loop gives us access to each element's key within a collection. With an array, this is each element's index. This isn't so useful.
```
droberts-ada: Since the `for...of` loop is more immediately useful / familiar from the Ruby curriculum, it might make sense to start with that.

I also think this section on for/in could use a little re-wording, to approach it from "where is it useful" as opposed to "where is it not useful". Maybe something like:

The `for...in` loop iterates over the keys of a collection. This is most useful when working with an object (similar to a Ruby hash). We haven't discussed objects in detail yet, but the following syntax should be familiar:

```js
let pets = {
  kari: "Kylo",
  dan: "Gecky",
  jamie: "Octo and Amper"
};
for (let human in pets) {
  console.log(`${ human }'s pets: ${ pets[human] }`);
}
// kari
// dan
// jamie
```

**Question:** What do you get when you use a `for...in` loop on an array? Why might this be useful?

10-JavaScript/exercises/ruby-to-js-worksheet.md
kariabancroft: Yes - felt like they were doing unnecessary work as almost exactly the same questions

10-JavaScript/exercises/functions-worksheet.md
kariabancroft: using a `class`? I'd like to review when this worksheet is assigned to see what we can/should be covering

10-JavaScript/welcome-to-js.md
kariabancroft: Ok - I was trying not to make this any longer than it needs to be, but I think this example is great!

10-JavaScript/exercises/objects-worksheet.md
```diff
@@ -10,7 +11,7 @@ console.log(pen.ink);
 2. What does the following code print to the console?
 ```javascript
 let bottle = {
-  contents: function () { return "some fine juice" },
+  contents: () => "some fine juice",
```
CheezItMan: Just an observation, arrow functions typically make poor methods.  This one doesn't matter since it doesn't use `this`, but we don't want students imitating it.

10-JavaScript/welcome-to-js.md
```diff
@@ -243,6 +243,45 @@ To point out the three components:
...
+```
+
+**Question:** What do you get when you use a `for...in` loop on an array? Why might this be useful?
+
+**`for`...`of`...**
+
+This `for` loop gives us access to each element's value within an array. This is the closest option to our `each` loop in Ruby, for this data type.
+
+```javascript
+let animals = ["horse", "cat", "dog"];
```
CheezItMan: Better to use single quotes.

10-JavaScript/exercises/objects-worksheet.md
kariabancroft: ok - i'll leave it now for a demonstration but noted.

10-JavaScript/welcome-to-js.md
kariabancroft: updated

PR #337, opened 2017-11-17
create resource template

kariabancroft:
## Description
As i'm developing new curriculum I think it would be good to templatize each resource. I just pulled what I think most resources have, but happy for feedback on adding/removing sections.

## Todos
none

## Feedback Requested
see above

### DIFF
--- a/doc-template.md
+++ b/doc-template.md
@@ -8,8 +8,8 @@
 | :------------- | :------------- |
 | Item One       | Item Two       |
 
-## Key Takeaways
-- sample
+## Key Takeaway
+Description of one or two sentences with key takeaway.
 
 ## Additional Resources
 - [sample]()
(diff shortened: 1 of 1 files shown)


kariabancroft:
[Textbook resource template](https://trello.com/c/gS4PDJVj/534-textbook-resource-template)

Hamled:
Could we also have a small example resource using each of the template's sections? I'm not sure I get what Key Takeaways should have vs. Learning Goals.

kariabancroft:
@Hamled I was referencing some of the HTML/CSS material we have like https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/05-html-css/layout-positioning.md

Should I add this link to the textbook?

Hamled:
It's probably not worth linking, especially if we end up having all of our textbook resources fit this template. Then there will be plenty of examples for what content each section should have.

PR #339, opened 2017-11-20
C8 Backbone Changes

droberts-ada:
## Description
This is a first pass - there's probably a fair bit of work still to be done.

High level philosophy: instead of forcing a bajillion random topics into the first week, let's dig deep on models and collections. These solve the most immediately obvious problem students will be facing (data organization) and will lead us to the next big problem (tying individual DOM elements to individual models).

See the README for the specific schedule. Basically Monday is "how do you even backbone", Tuesday is "OK but why wouldn't I just use arrays", and Wednesday is anything that doesn't fit in the previous two days.

The livecode for week 1 is called BackBooks, and is a SPA version of the rails library we built long ago. This ends up making more sense than TaskList, for two reasons:
- TaskList implies the ability to modify / delete individual tasks, which we won't get until we do views
- A book has several fields you might sort on, which is key for one of these lessons

## Other Repos

The end result of the live code lives at https://github.com/AdaGold/backbooks-client/tree/dpr/library. There is still a lot of work to do here (build a baseline and a reference for the end of each lecture).

The companion Rails API is 100% complete, and lives at https://github.com/AdaGold/backbooks-rails-api

## Todos
- [x] Clean up the `backbooks-client` repo
    - [x] Add baseline branch with nothing on it (probably master)
    - [x] Add branches for each step of the way
- [ ] Move codepens (particularly in the underscore lecture) to the shared Ada account
- [ ] Build Wednesday's "Advanced Models" lecture
- [ ] Do some work organizing on Week 2, since that also will be very different. Probably keeping the TaskList live code here.
- [ ] Build the new week 1 project (BackTREK)
- [ ] Remove all the old lectures, instead of just renaming them `-old`
- [ ] Lots of diagrams
- [ ] Finish all the `TODO DPR`s I left (mostly covered above, but do a final pass with grep to be sure)
- [ ] Final pass on the README
- [ ] Find a good resource on Backbone Events that doesn't require knowledge of views

## Feedback Requested
- Find my dumb typos!
- I've tried to include many moments where we pause and re-engage students, either by asking comprehension questions, by having them do the same thing again with different nouns, or by having them forge ahead and write some component in small groups. Are there enough of these (especially in later lectures)? Are they targeted enough? Suggestions here are welcome.
- Check links to worked versions - do they all point to the right places?


11-Backbonejs/events.md
```diff
@@ -0,0 +1,207 @@
...
+TODO DPR: expanded diagram similar to above but w/ event names
+
+1. A user fills out and submits the "Add a book" form, triggering a `submit` DOM event
+1. Our `submit` event handler will read the form data and use it to add a book to the collection
+1. When the book is added, the collection will emit a `update` event
+1. Our `update` event handler will re-build the table of books in the DOM
+
+We will build these components in reverse order, starting with the handler for the `update` event, and then adding code to read the form and add a model to the collection.
+
+### Building the `update` Event Handler
```
kariabancroft: This section feels like a wall of text with very little action - consider how we might consolidate this

11-Backbonejs/events.md
```diff
@@ -0,0 +1,207 @@
...
+bookList.on('update', render);
+```
+
+**Question:** Where should this registration code be placed? Does it need to be inside `$(document).ready`? Why or why not?
+
+**Question:** How would we test this code manually, to ensure we've connected all the pieces correctly?
+
+### Triggering the `update` Event
+
+Now that we've got our event handler in place, we need to build some code that triggers an `update` event by adding a model to our collection. This code itself will be an event handler, waiting for the `submit` even on the form.
```
kariabancroft: "event" typo

11-Backbonejs/events.md
```diff
@@ -0,0 +1,207 @@
...
+
+Working with the person next to you, build this logic. Things to think about include:
+
+- Where will you define this function?
+- How will you read the values from the form?
+- How will you turn these values into a new model in the collection?
```
kariabancroft: model instance or model object

11-Backbonejs/events.md
```diff
@@ -0,0 +1,207 @@
...
+
+Work with the person next to you to write code to detect a click event on the `<th>` elements and sort the collection by the appropriate attribute.
+
+The handler will need to set `bookList.comparator` to the name of the field, and then call `bookList.sort()`.
+
+There are three columns we need to be able to sort by, so this code will be a little complex. Start by building a click handler for only the `title` field. Once you've got that working, generalize your code to handle all three fields. Do your best to keep things DRY.
+
+We'll come back as a class to review this code. Once you're finished, you should have something like this.
+
+Once you're done, you should have something that resembles the [code on the `sort` branch](https://github.com/AdaGold/backbooks-client/blob/sort/src/app.js).
```
kariabancroft: Just a dupe line to the above.

11-Backbonejs/events.md
```diff
@@ -0,0 +1,207 @@
+# Working with Backbone Events
```
kariabancroft: It seems like this could potentially be two different resources. One that explains how events work overall (with the vocab and diagrams) and then a second with the implementation examples and exercises. Thoughts?

11-Backbonejs/Backbone-APIs.md
```diff
@@ -7,13 +7,11 @@ By the end of this lesson you should be able to...
 - Send Backbone data to an API
```
kariabancroft: Use Backbone to send data to an API

11-Backbonejs/Backbone-APIs.md
```diff
@@ -23,101 +21,115 @@ Note that these instructions will work just as well for a deployed API - just ch
...
-2. read the documentation
-3. apply some changes to the code
-4. test it out
-5. rinse & repeat
+From our reading of the API earlier we know that the Rails `Book` model has validations, which means our API call will fail if the user sends bad data. We know from experience how important it is to report errors to the user in a polite and actionable way. This will both improve the user experience and make developing the application much less painful.
 
-<!-- We're going to pass over "Edit" in this lecture, because the way we implemented it doesn't match up with an API well. It's worth spending some time looking at our current implementation to figure out why it will be difficult, and thinking about how we could have implemented it better. -->
+Fortunately, Backbone allows you to define success and error callbacks in the arguments to `.save()` (and `.fetch()`).
+
+One tricky bit of syntax: the hash specifying these callbacks must be the second argument to `.save()`. The first is reserved for if you want to specify which attributes to send. If you don't specify them Backbone will send everything, which is what we want, so add an empty hash as the first argument.
```
kariabancroft: The JS object (not hash)

10-JavaScript/underscore-templates.md
```diff
@@ -125,34 +124,34 @@ First, notice that we've put this bit of code in `$(document).ready`. Even thoug
...
+  let generatedHtml = todoTemplate(todoData[i]);
   $('#todo-list').append($(generatedHtml));
 }
 ```
 
-Note that Underscore doesn't assume we have access to jQuery. That means the return value from the template is just a vanilla JavaScript DOM object. If you want to do fancy jQuery things with it, you'll have to pass it through `$`, like we do here.
+Note that Underscore doesn't assume we have access to jQuery. That means the return value from the template is a vanilla JavaScript DOM object. If you want to do fancy jQuery things with it, you'll have to pass it through `$`, like we do here.
 
 ### Putting It All Together
 [Here is a CodePen](http://codepen.io/droberts-ada/pen/wodpWe?editors=1011) that does all the things we just talked about!
```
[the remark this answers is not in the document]
droberts-ada: Good catch!

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Defining The Collection
+
+1. Create a file `src/collections/book_list.js`
+1. Open that file in Atom, and add the following code:
+    ```javascript
+    import Backbone from 'backbone';
+    import Book from '../models/book';
+
+    const BookList = Backbone.Collection.extend({
```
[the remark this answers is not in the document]
droberts-ada: I did not know this was true. Do you have a source on this? I can't find anything about it in the Backbone docs, and people on the internet seem to do all sorts of things.

11-Backbonejs/models.md
```diff
@@ -0,0 +1,148 @@
...
+import Book from './models/book'
+```
+
+Because the path begins with a `./`, Webpack will interpret this as a relative path, which should get us to `src/models/book.js`. It will load whatever that file `export`s, in this case the `Book` constructor, and make it available in an object called `Book`.
+
+**Exercise:** Working with a whiteboard and the person next to you, write the code to define a `Cat` model.
+
+#### Creating Instances
+
+To create instances of `Book`, invoke the `Book` constructor with the `new` keyword. Pass in a hash with the attributes you want.
```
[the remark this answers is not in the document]
droberts-ada: I have added a section on this, but I feel this makes the "defining a model" section even more of a wall of text. Not sure what to do about this.

11-Backbonejs/models.md
```diff
@@ -0,0 +1,148 @@
...
+
+```javascript
+let book = new Book({
+  title: "Ancillary Justice",
+  author: "Ann Leckie",
+  publication_year: 2013
```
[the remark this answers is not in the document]
droberts-ada: Question: for a good chunk of this live code we get the data from a Rails API, where the naming convention IS to use snake case. How do we balance this?

11-Backbonejs/README.md
```diff
@@ -1,14 +1,36 @@
 
 # Backbone
 
-## Part 1
+## New order
+### Monday
+1. [Intro to Backbone](Introduction-to-Backbonejs.md)
+1. [Underscore Templates](../10-JavaScript/underscore-templates.md)
```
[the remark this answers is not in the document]
droberts-ada: Yeah, I like it.

11-Backbonejs/events.md
droberts-ada: I think the theory one would not be big enough to justify a whole separate resource.

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
...
+
+// Get a subset of books matching a condition (similar to Ruby's select)
+let recentBooks = bookList.filter((book) => {
+  book.get('publication_year') > 1990
+});
+```
+
+Note that collections implement several iterators that ES6 arrays do not, that most of the iterators have an alias (`forEach` -> `each`, `reduce` -> `collect`, etc.) and that the names on some of the more obscure iterators may differ. [You can find a complete list in the Backbone documentation](http://backbonejs.org/#Collection-Underscore-Methods).
+
+This discrepency exists because Backbone predates ES6 by several years. In fact Underscore's iteration functions (which Backbone uses) were a big part of the inspiration for ES6.
```
CheezItMan: discrepency ==> discrepancy

11-Backbonejs/collections.md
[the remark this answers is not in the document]
kariabancroft: These two refs:
https://stackoverflow.com/questions/8695254/what-are-some-backbone-js-naming-conventions
http://codebeerstartups.com/2012/12/8-collections-in-backbone-js-learning-backbone-js/

11-Backbonejs/models.md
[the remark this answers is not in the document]
kariabancroft: Can this be part of the `fetch` implementation? Otherwise, I guess we can leave it.

11-Backbonejs/models.md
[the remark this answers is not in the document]
kariabancroft: It seems good with the "no schema" section. I like it.

11-Backbonejs/events.md
```diff
@@ -29,12 +29,8 @@ Once this code has been run, whenever our `bookList` emits a `update` event the
 
 Why is this important? **It means we can separate code that updates the model from code that updates the DOM.** In practice, this allows us to simplify and DRY our code that handles DOM events.
 
-TODO DPR: diagram. Something like
-```
-     => DOM event handler (jQuery) =>
-DOM                                    Model / Collection
-     <= Backbone event handler     <=
-```
+![backbone-event-cycle](images/model-event-diagram-generic.png)
```
kariabancroft: Do we have the resource from which this diagram was created when/if we need to edit? I like to add them as a comment after the image

10-JavaScript/underscore-templates.md
```diff
@@ -25,7 +25,7 @@ There are three steps to Underscore templating:
 The data we'll be working with is a TODO list. Here is some sample data:
 
 ```javascript
-let todoData = [
```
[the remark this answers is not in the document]
droberts-ada: No - I intentionally left it as-is, to reinforce the fact that underscore templates are not actually a part of backbone.

11-Backbonejs/collections.md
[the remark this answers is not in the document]
droberts-ada: If I have time to address this, it will be as a separate PR / change to the livecode. However, propagating it through all those branches will be a lot of work potentially.

11-Backbonejs/events.md
droberts-ada: Chris did the work, I have asked him for it.

11-Backbonejs/models.md
```diff
@@ -0,0 +1,159 @@
...
+Because the path begins with a `./`, Webpack will interpret this as a relative path, which should get us to `src/models/book.js`. It will load whatever that file `export`s, in this case the `Book` constructor, and make it available in an object called `Book`.
+
+**Exercise:** Working with a whiteboard and the person next to you, write the code to define a `Cat` model.
+
+#### Creating Instances
+
+To create instances of `Book`, invoke the `Book` constructor with the `new` keyword. Pass in a hash with the attributes you want.
+
+```javascript
+let book = new Book({
```
Hamled: `let` -> `const`

11-Backbonejs/models.md
```diff
@@ -0,0 +1,159 @@
...
+
+```javascript
+console.log(book.attributes);
+// => { title: 'Ancillary Justice', author: 'Ann Leckie', publication_year: 2013 }
+```
+
+**Question:** How would we get the information in an instance of `Book` to show up in our web page?
+
+```javascript
+let bookHTML = bookTemplate(book.attributes);
```
Hamled: `let` -> `const`

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
+# Organizing Data with Collections
+
+As we've worked with JavaScript, up until this point all our data in has been stored as arrays of objects. This is fine for simple programs, but for more complex applications we'll need a more robust way to organize things. In the Backbone framework, this problem is solved through Models and Collections.
```
Hamled: This paragraph is the exact same text we have in `models.md`, is that intentional?

If we want to keep it we should fix "all our data in" -> "all our data".

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
...
+
+Now that we've defined our `Book` Model, it's time to build a Collection to contain it. We will call this Collection `BookList`.
+
+### Defining The Collection
+
+1. Create a file `src/collections/book_list.js`
+1. Open that file in Atom, and add the following code:
+    ```javascript
+    import Backbone from 'backbone';
+    import Book from '../models/book';
```
Hamled: This can simplified to:
```javascript
import Book from 'models/book';
```

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+#### Creating a Collection
+
+To create an instance of `BookList`, use `new BookList()`. If you pass in no arguments, the list will be created empty of books.
+
+A `BookList` can also be created with books. If you pass an array to `new BookList`, Backbone will attempt to create a model instance from each element in the array, using whatever model you specified when you defined the collection. For this we can use the `rawBookData` variable that is already defined.
+
+```javascript
+let bookList = new BookList(rawBookData);
```
Hamled: `let` -> `const`

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
...
+**Exercise:** On the whiteboard, create a new `CatList` with two cats.
+
+#### Adding Books
+
+You can also add models to an existing collection. There are two ways to do so.
+
+If you already have a model, you can use the `.add` function to add it to a collection:
+
+```javascript
+let book = new Book({
```
Hamled: `let` -> `const`

11-Backbonejs/collections.md
```diff
@@ -0,0 +1,164 @@
...
+Backbone Collections provide a large number of methods to iterate through their elements. For the most part these look similar to the ES6 array iteration functions you already know and love. For example, `forEach`, `map`, `filter` and `reduce` are all included.
+
+```javascript
+// Log a description of each book
+bookList.forEach((book) => {
+  console.log(`${ book.get('title') } by ${ book.get('author') }`);
+});
+
+// Get a subset of books matching a condition (similar to Ruby's select)
+let recentBooks = bookList.filter((book) => {
```
Hamled: `let` -> `const`

PR #344, opened 2017-11-22
Cleanup the Backbone section's README

Hamled:
## Description
* Took out the "old order" section because we don't need to have it alongside the new stuff.
* Also took out the week 2 section, we should add it back with whatever PR actually updates those curriculum items, or when the schedule is determined.

**NOTE**: Chris' updates (#341) for advanced models & collections curriculum also updates `README.md`, so we should get that PR merged in first and then I can fix any resulting conflicts for this PR afterwards.

## Todos
- [ ] Consider whether we should really be breaking the curriculum items into specific days... we've been burned by making our organization info too schedule-specific before.

### DIFF
--- a/11-Backbonejs/README.md
+++ b/11-Backbonejs/README.md
@@ -15,9 +15,3 @@
 
 ### Wednesday
 1. Advanced Model stuff (TODO DPR)
-
-### Week 2
-1. Views
-1. Custom Events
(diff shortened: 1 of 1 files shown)


Hamled:
Closing this because the work here was already done in b64e7bf.

PR #346, opened 2017-11-23
Backbone Models curriculum cleanup

Hamled:
## Description
I reworded the instructions for how to do `import` statements with our project baseline, because we do not need to do relative paths when importing from within the `src/` directory.

Also changed `let` to `const` in code snippets and made some minor grammar changes.


droberts-ada:
I tried using relative paths without './' with the webpack setup and wasn't able to get it to work. It could be I'm misremembering, or that I forked too early and missed some critical piece of infrastructure.

Have you tried this in the live code repo to confirm it will work?

droberts-ada:
Also, if we're updating code here we'll want to go back and update all the branches in the live code to match. My branching strategy was pretty silly (basically, build the whole finished thing and then remove pieces), so this might be a giant pain.

droberts-ada:
OK, it appears Chris made changes to the baseline (particularly fixing the search path for loading files) after I forked. I will go back and make sure everything is in sync.

PR #357, opened 2018-01-05
Initial React

kariabancroft:
## Description
Overall initial concepts and structures for the first few React lectures. I feel pretty good about the order of the lectures, but still want to flesh things out a bit more. I think we have enough here for a first basic project, but would potentially like confirmation on that.

## Todos
- [x] Do a better job with the context for the "creating components" lecture - give more structure on what/why we're adding another component
- [x] Cleanup some "additional resources" and "key takeaways sections"
- [x] Add a separate "practical application" of state vs props lecture which extends the concepts and applies them to the basic hello world application that we've been creating/modifying

Not as part of this bundle of lectures but future topics will include:
- Eventing
- Data validation using propTypes
- API integration
- Forms
- Router

## Feedback Requested
- Does the overall order initially look good? 
- Do I have any huge gaps or errors in existing examples?
- More diagram opportunities?


droberts-ada:
[React Curriculum](https://trello.com/c/75yiTyDM/755-react-curriculum)

React/creating-components.md
```diff
@@ -0,0 +1,79 @@
...
+
+Each component will be a small, reusable piece of code that manages their own **content, presentation and behavior**. 
+
+### Overview
+It is important to understand what is happening currently within the React application so we can model the same pattern with a custom component. Right now, our React application renders the `App` component from the `index.js` file and puts it in the DOM element where the ID is "root". Normally, this is only done **once** in a React application, so we are going to leave this code alone as we build on.
+
+Next, our `App` component contains a render function will which render a `header` and a `p` using JSX.
+
+The powerful thing about React is that **any component can render any other component**. That is, we can (and should) **nest our components** within one another to create the most effective React application design.
```
Hamled: We touch upon the idea of nesting components in a few spots in this textbook resource (including talking about it as the key takeaway), however I don't feel like it's really getting sufficient emphasis. Especially since the next activity asks them to break down a website into a hierarchy of nested components.

We should at the very least make it clear that nesting components isn't just a powerful aspect of React's design, but is a necessary structure for modelling a web page. Because web pages are about representing a physical space (a flat 2D plane, like a piece of paper), it is possible to consider every location/element within that space as being contained in one or more larger spaces -- much like a map of the US has states, and within them has counties, and cities, and streets, etc.

We could also make the comparison to HTML itself, which forces a hierarchical/nested structure in the language itself. Every element on the page in a React app is the result of a single component's `render()` function, and because elements in HTML are necessarily contained inside of other elements, our components must be nested in a way that reflects the same structure as the HTML.

React/creating-components.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Context
+As we continue using this application, we are going to be creating a React application that manages our list of Ada students! We are going to use different components to handle different portions of our student listing.
```
Hamled: I know that you've already got a lot of curriculum content written for building a list/database of Ada students... but are there any other reasons we shouldn't stick with the books examples from the Rails curriculum?

React/creating-components.md
```diff
@@ -0,0 +1,79 @@
...
+1. Export the class at the bottom of the file. Note that whatever is exported here needs to match the name of the class created in the step above. This will allow us to import this component to use in other files.
+    ```JavaScript
+    export default Student;
+    ```
+
+1. **QUICK BREAK!** Check in with your seat squad to verify that they are at the same place in setting up their component file. Look at [the code](https://github.com/AdaGold/react-hello-world/blob/part-1/src/components/student.js) if you need to copy/paste to catch up.
+
+### Customize our Component
+
+With your seat squad, add some JSX elements to a `render()` function that you add to this new component. We can't fully test it out quite yet, but you can make sure that your React application compiles and does not show you any errors in the browser as confirmation of success. Feel free to hard-code some student data and HTML that you think might be relevant.
```
Hamled: I'm certainly curious to see how this turns out. Up until this point the students have had very little opportunity to try writing their own JSX or `render()` function...

Also, until they have an `import` line for the new Student component's file, React will not attempt to compile this file and won't give any errors if there are syntax issues. Should we briefly go over the terminal output from `npm start` (or the browser error display) and how to identify when there are compilation errors?

React/creating-components.md
```diff
@@ -0,0 +1,79 @@
...
+    import Student from './components/student';
+    ```
+
+1. Now we can render our `Student` component within our App component. Add `<Student />` to the existing render function in the App component. Note that the render function can only return one outermost element, so the new component must be **nested within** that outermost element.
+
+1. Verify! Run `npm start` to take a look at your application running in the browser and ensure that you see the elements you added to your student component.
+
+If you need to, take a look at this final version of [the code](https://github.com/AdaGold/react-hello-world/tree/part-2/src) to verify your work.
+
+** What have we accomplished so far?**
```
Hamled: This is not being rendered in bold, possibly due to the space before "What"?

React/exercises/identifying-components.md
```diff
@@ -0,0 +1,13 @@
+# Identifying Components
+In this exercise, we will go through a popular website and brainstorm how to break down the web application into components.
+
+To get you started, we've done an example of this using the MediaRanker application that you're already familiar with.
+
+![Media Ranker as Components](../images/media-ranker-components.png)
```
Hamled: This is pretty minor, but the image suggests that the Navigation component contains the MediaSplash component (and thus all other parts of the page except the logo). This doesn't seem like a normal React component setup.

React/creating-components.md
```diff
@@ -0,0 +1,79 @@
...
+1. Create a new file in this new folder called `student.js`
+
+1. In this new file, import the React and Component libraries:
+    ```javascript
+    import React, { Component } from 'react';
+    ```
+
+1. Create the class definition with the name of the component you want to create. Note that this does not need to match the file name.
+    ```javascript
+    class Student extends Component {
```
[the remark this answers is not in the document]
droberts-ada: I would probably not tell them that the component name and the file name don't have to match. As far as I'm concerned, they definitely _should_ match.

React/exercises/identifying-components.md
```diff
@@ -0,0 +1,13 @@
...
+
+Break into pairs and take a look at CraigsList. Identify the visual "components" the website is comprised of. We suggest using using your white board or paper to draw these out on your table! So something like this...
+
```
droberts-ada: Something like what?

React/creating-components.md
droberts-ada: It might be helpful to the global learners to include a link here to a completed version of the app, so that we know what we're working toward.

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+## Learning Goals
+- Continue nesting components within one another
+- Continue practice component creation and JSX syntax
+
+Next we are going to extend our application to support multiple students. Additionally, we are going to support components dynamically loading data.
+
+## Load data from `props`
+
+First, we want to use the Student component we already created, but we want to allow this to load data dynamically.
+
```
droberts-ada: I think this exercise would work well in the `props and state` lesson.

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+Once the prop names are **passed in** to the component, we can then use them within the component. That's where `this.props` comes in!
+
+## Setup to manage list of data
+One student is great and all, but ideally, we'd really be able to manage our whole list of students that we might need.
+
+We'll create a new component called `StudentCollection` that will live between the existing  `App` and `Student` components. This new component will manage the full list of students and then render the data for each student.
+
+Spend a few minutes now going back to the [creating components](creating-components.md) notes to see how to create the new component file with the basic class, imports and exports you need.
+
```
droberts-ada: I like that you call back to previous learning and make them do it on their own!

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+    this.state = {
+      students: [
+        {
+          name: "State Ada",
+          email: "state-ada@ada.co"
+        },
+        {
+          name: "Cool",
+          email: "cool@ada.co"
```
droberts-ada: I like the idea of changing the student names every time so they know for sure that they're getting the new version. However, naming a student "cool" feels a little silly. I think that's fine if you need to make up data on the spot, but in the official notes it's a distraction.

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+
+
+## Use `state` for data
+
+Set up a constructor in the `StudentCollection` component which will use a variable to create a state-managed collection of student information. Don't forget your call to `super();` which is always required in a component constructor!
+
+```javascript
+constructor() {
+    super();
+
```
droberts-ada: Could you add comments to the tops of these code blocks telling us which file we're currently in?

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+Note that there are a few different ways to accomplish this same result. The approach we've taken here is to create a new variable _outside_ the `return` to construct the individual components. Then, within the return, we wrap the variable from above in a single `div` since we can only return one high-level element.
+
+Verify that you now see the data from the constructor in your browser. If you need to see what has gone wrong, take a look at [the code](https://github.com/AdaGold/react-hello-world/blob/part-5/src/components/student_collection.js).
+
+**Dive In:** Unpack the line with `return <Student...` with the person sitting next to you. Identify each piece of code on this line. Some questions to think about as you go through this activity:
+- What do you think the `studentComponents` variable looks like after this code executes?
+- How are `props` used within this code?
+- How does this code relate to the code within the `Student` component?
```
droberts-ada: I like this list of questions. Maybe add "what is being `return`ed, and where is it going?"

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+    <div className="student">
+    ...
+  );
+}
+```
+
+We don't want to add the class name onto each **component** that is rendered from the `StudentCollection` because that wouldn't go directly to the HTML. We should instead add it directly to the HTML **within** the component.
+
+You can see the final version of the code [here](https://github.com/AdaGold/react-hello-world/blob/part-6/src).)
+
```
droberts-ada: Dangling paren on line 132

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+## Some CSS
+We want our page to have a nice look & feel! Let's add some simple CSS.
+
+We can make use of the existing css file that comes from within our boilerplate React application. Let's add a new class in the `App.css` file for the student.
+
+```css
+.student {
+  padding: 10px;
+}
```
droberts-ada: Maybe make this something more visually distinctive, so that it's immediately obvious whether or not it worked. Something like:

```css
.student {
  width: 15em;
  border: 1px solid black;
  background-color: hsl(189, 100%, 89%);
  padding: 1em;
  margin: 1em;
}

.student-list {
  display: flex;
  justify-content: flex-start;
  flex-wrap: wrap;
}
```

(you would have to put the `.student-list` class on a div wrapping all the students but not the `<h3>` in the collection)

React/lifecycle-methods.md
```diff
@@ -0,0 +1,82 @@
+# Lifecycle Methods
+
+## Learning Goals
+- Learn about the three phases: mount, update and unmount
```
droberts-ada: I like this lesson. Looks like this will be followed by talking about APIs the same day, right? I think that's important for giving them a reason to care.

React/component-design.md
```diff
@@ -0,0 +1,40 @@
...
+
+These are used to simplify a component when it's only purpose is to render HTML or other nested components. The definition of these components use a simplified arrow-function syntax since they don't need to be a fully functional `class` like other components.
+
+```JavaScript
+const MyFunctionalComponent = (props) => {
+  return <h1>This is just a header with this name: {props.name}</h1>;
+}
+```
+
+Notice in the above example we can still pass in props into this component, but we access that variable directly from the parameter passed in, rather than through `this.props` like we would with a class-based React component.
```
droberts-ada: Can you have them immediately apply this to the livecode? Maybe the `<Student>` component, unless some other yet-to-be-written lecture makes it stateful.

React/component-design.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Container Components with Presentational Components
+This design pattern provides us with a common way to organize components within one another. Container components will manage the creation and data sent to presentation components. Presentation components will be solely responsible for displaying data, and not for managing state within themselves. Because of this, these presentation components are often the same as functional stateless components.
+
+One way that this pattern can come about in our React applications, is by using the **container** component to manage **state**, and having two (child) **presentation** components, one which manages **data input and modification** and another which manages the **presentation of this data**.
+
+![stateful components using stateless components](images/stateful-stateless-components.png)
```
droberts-ada: I like the diagram, but I think they're going to need an immediate application in the livecode or they will just not care about this.

React/creating-components.md
kariabancroft: Yes - they are _definitely_ sick of the books examples after Rails

React/props-state.md
```diff
@@ -0,0 +1,99 @@
...
+Here is a helpful chart to assist you in determining whether data belongs in `props` or `state`. For any line item that has "Yes" for both, it means you need to make that decision based on the context of the problem you are trying to solve.
+
+ ?     | props     | state
+ :------------- | :------------- |:-------------
+Can get initial value from parent component?       | Yes   | Yes
+Can be changed by parent component? | Yes | No
+Can change inside component? | No | Yes
+Can set initial value for a child component? | Yes | Yes
+Can change in a child component? | Yes | No
+
```
tildeee: This material is great imo, especially because it stands alone. I definitely like the emphasis of "here's a lot of definition and you'll refer to these notes later"

It of course ties directly into the next lecture, but it definitely feels weird without the application. I wish there were a way to address this, though I can't think of any suggestions right now

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+    - This shouldn't change anything yet in our application, so save the file and refresh the browser to verify that the data still looks like the original hard-coded data.
+
+1. Update the `render` function in the `Student` component to replace the hard-coded values with code tags `{ }` which contains the `prop` that came in from the parent component.  
+    - It should now contain `{ this.props.name }` and `{ this.props.email }`
+    - Verify that the content displayed is now coming from the values in the `App` component.
+    - If you need to verify, take a look at [the code](https://github.com/AdaGold/react-hello-world/tree/part-3/src) so far.
+
+
+If we identify each individual piece of the component rendered, we'll see:
+![component prop breakdown](images/component-prop-breakdown.png)
```
tildeee: this is the perfectest diagram ever and I wish we had things like this for the beginning Ruby section when students are still learning how values get assigned to variables which get passed around

React/creating-components-2.md
```diff
@@ -0,0 +1,138 @@
...
+
+**Dive In:** Unpack the line with `return <Student...` with the person sitting next to you. Identify each piece of code on this line. Some questions to think about as you go through this activity:
+- What do you think the `studentComponents` variable looks like after this code executes?
+- How are `props` used within this code?
+- How does this code relate to the code within the `Student` component?
+- What would happen to the HTML rendered if you remove either `name` or `email`?
+
+Now let's take a look at an updated version of the diagram that we created in our last component creation lecture:
+
+![nested components](images/nested-components.png)
```
tildeee: I could see this diagram being pulled up to the section where we introduce the idea of a `StudentCollection` so the students know how this component is supposed to fit in (and also being repeated here)

React/README.md
```diff
@@ -0,0 +1,10 @@
+# React
+
+1. [Intro to React](intro-to-react.md)
+1. [Hello World](react-hello-world.md)
+    - [Exercise](exercises/intro-to-react.md)
+1. [Creating Components](creating-components.md)
+1. [`props` & `state`](props-state.md)
+1. [Creating Components Part 2](creating-components-2.md)
+1. [Lifecycle Methods](lifecycle-methods.md)
+1. [Component Design](component-design.md)
```
tildeee: I know everything's in flux, but this order (Lifecycle Methods before Component Design) doesn't match the proposed calendar

React/props-state.md
```diff
@@ -0,0 +1,99 @@
...
+## Overview
+Props & State are the two ways that React components manage and share their data.
+
+## `props`
+Props are **passed from the outside of the component** into the component itself. Props are accessible using `this.props` within a component. Props cannot be changed from within the component who received the prop values.
+
+#### Overview
+- Can be assigned to any variable name you want
+- Can contain any object or function you would like
+
```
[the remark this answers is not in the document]
kariabancroft: neither - I was thinking about what the variable name is when the prop is passed in: `<MyComponent myVal={3 * 3} />` or something.

React/props-state.md
```diff
@@ -0,0 +1,99 @@
...
+## `state`
+State is managed **within a given component**. State can be accessed using `this.state` and it can be modified using `this.setState()`.
+
+#### Overview
+- Can be assigned to any object key you want
+- `setState` will always take an object as the parameter
+- `setState` will merge any params passed in with the existing state data, it won't overwrite it
+- We often set the initial state in the constructor function
+
+A really important aspect of state is that whenever the state of a component is changed, it will be **updated**, calling the `render` function. For now, you can think of `setState` as doing the `setState` operation PLUS the `render` _automatically_. We'll learn even more about what else our component does when it is updated later on.
```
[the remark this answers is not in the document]
kariabancroft: Are you referring to an instance variable in Ruby? I feel like they're not the same because they're not directly accessible or changeable in the way that instance variables are either. State should never be directly accessed externally

React/creating-components-2.md
kariabancroft: Great suggestion

PR #358, opened 2018-01-09
Updates to Initial OOP Curriculum for C9

droberts-ada:
## Description
- Split resource into a short theory section (good for pre-reading or a group lecture) and a practicum
- Rework the example class from the practicum to `User` instead of `Chair`
- Add a short discussion of the `self` keyword

Splitting things up is the first commit, so if you want to see individual change-sets you'll have to go commit-by-commit. Sorry.

## Todos
N/A

## Feedback Requested
- I have run all the code examples and made sure they work, but if I missed a `Chair` somewhere in the text let me know.
- The section on `self` is pretty minimal - it's important to know it exists, but not super useful yet. Can you think of a better way to present it?
- The practicum is still pretty long, maybe it needs to be shortened or adjusted.


droberts-ada:
[C9 Week 1: Work on initial OOP resource](https://trello.com/c/Iq0tkVHx/591-c9-week-1-work-on-initial-oop-resource)

droberts-ada:
It looks like this will land on Thursday of the first week, which means this probably won't be an auditorium lecture. We might do the theory section as a "read this at home".

I do agree that there's a lot of vocab here, and that it would be worthwhile reinforcing that as much as possible. I will spend some time thinking about how to do this.

Looking at it, I'm noticing that the theory section is also missing a summary, so I'll go ahead and add that.

01-ruby-fundamentals/classes-overview.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Question:** what _state_ and _behavior_ might a `Book` class need?
+
+Note that all this state and behavior is for a particular instance of a `Book`. Think about the abstract concept of a book, what is its title or author? That doesn't make any sense. Only when we create an instance do these properties come into focus.
+
+When modeling a real-world object like a book or a person, it's important to remember that you don't need to track everything about it, only what's important to your program. You certainly could include information about the color of the book's cover or how tall it is in your catalogue, but you should only do so if that information is useful to you.
+
+Different kinds of objects have different state and behaviors, and the same real-world object in two different programs might have totally different representations. Think about how different a model of a person would be in a medical application versus in a gradebook. What information is relevant is totally different for the two programs.
```
CheezItMan: Last sentence needs a question mark.

droberts-ada: I don't think it does. Maybe it could be reworded to not start with a question word.

PR #362, opened 2018-01-22
Update the Hashes and Arrays Resource

droberts-ada:
## Description
Restructure the Arrays and Hashes lecture

There are now 3 main sections: Arrays, Hashes, and In Combination. Each section has the same basic layout:
- Creation
- Accessing elements
- Iteration

which I think gives it a really nice flow and emphasizes the pattern.

In between arrays / hashes and in combination is a comparison section, where I moved the table I made in the previous commit.

I've also reworked the examples away from Pokemon, since we got so many freaking complaints last time.

We may want to assign the first 3 sections (arrays, hashes, comparison) as homework the night before, depending on time.

## Todos
- [x] Build the pre-reading into the calendar

## Feedback Requested
Find my dopey typos!


droberts-ada:
[C9 Week 1: Consolidate initial treatment of Arrays and Hashes](https://trello.com/c/Gy1TeXm2/587-c9-week-1-consolidate-initial-treatment-of-arrays-and-hashes)

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
-An array is a list of data and each item in the list can be accessed or *indexed* by a number.  The first element of any Array is index number 0.  You can always access the size of the Array with the length method.  
+Here are some facts about arrays.
+
+#### Creation
+
+- Arrays are created by putting square brackets `[]` around a list of things
+  - The list can be empty
+- Since an array is an object, we can store it in a variable
+- You can get the length of the array with `Array#length`
+- The array will keep track of the order elements were added
```
CheezItMan: I think it's better, it maintains the original order of the elements given.

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
-A Hash is similar to an array in that it's a collection of data.  However to access an individual element in an Array you always use a number as a key to get to that element.
+Here are some facts about hashes.
+
+#### Creation
+
+- Hashes are created by putting curly braces `{}` around a list of `key: value` pairs
+  - The list can be empty
+- Since a hash is an object, we can store it in a variable
+- You can get the number of key/value pairs with `Hash#length`
+- The hash does **not** keep track of the order of its key/value pairs
```
CheezItMan: I think technically ruby does now maintain the order of the key/value pairs.

http://ruby-doc.org/core-2.5.0/Hash.html

I think it makes more sense to talk about why you shouldn't worry about order in a Hash.

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
+puts fruit_prices["orange"]
+# => nil
+# Why?
 ```
 
-One thing to remember is that the keys must be unique.  For example in the Hash above, you couldn't use "Squirtle" to refer to another value.
+#### Iteration
+
+- Like an array, you can iterate through the key/value pairs in a hash
+  - Unlike an array, the order is not guaranteed
```
CheezItMan: Again I think it is in the order they were inserted.

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
-| 2       |  250  |    550    |
-| 3       |  128  |    678    |
-| 4       |  64   |    742    |
-| 5       |  3016 |    3758   |
 
+#### Iteration
 
-This loop does the same thing with slightly different syntax.  
+- Iteration is the process of doing something to each element in the array, like printing it to the command line or adding 7 to it
+- To iterate manually, you can use a `times` loop on the length of the array
```
tildeee: What does "to iterate manually"?

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
-| 1       |  300  |    300    |
-| 2       |  250  |    550    |
-| 3       |  128  |    678    |
-| 4       |  64   |    742    |
-| 5       |  3016 |    3758   |
 
+#### Iteration
 
-This loop does the same thing with slightly different syntax.  
+- Iteration is the process of doing something to each element in the array, like printing it to the command line or adding 7 to it
```
kariabancroft: I might leave off the 'like printing it' section

01-ruby-fundamentals/data-structures.md
kariabancroft: Maybe describe what this does differently (using the index) rather than using "manually"

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
+- The hash does **not** keep track of the order of its key/value pairs
 
-So for this list of Pokemon:
 ```ruby
-my_list = [ "Bulbasaur", "Ivysaur", "Venusaur", "Charmander", "Charmeleon", "Charizard", "Squirtle", "Wartortle" ]
-```
-The Array has the following index/value pairs.
+fruit_prices = {
+  apple: 2.15,
+  pear: 3.02,
```
kariabancroft: Do we prefer the trailing comma? I feel like do in JS but not in Ruby

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
+menu = [
+  {
+    name: 'beet salad',
+    price: 6.75,
+    allergens: ['nuts',],
+  },
+  {
+    name: 'quiche',
+    price: 10.00,
+    allergens: ['gluten', 'dairy',],
```
kariabancroft: Same question re trailing commas

01-ruby-fundamentals/intro-to-object-types.md
```diff
@@ -197,13 +199,17 @@ ada.length
 # => 4
 ```
 
+
+
+
+
 Look at the Ruby docs for [Array](http://www.ruby-doc.org/core-2.4.0/Array.html) and [Hash](http://www.ruby-doc.org/core-2.4.0/Hash.html) for more details about collections.
```
kariabancroft: We should update these links to not reference the specific version but reference the latest by just using `/core/` instead

01-ruby-fundamentals/data-structures.md
droberts-ada: You're right, it's not mandated by the spec but it does happen in practice. I've struggled with this because ordering is **not** a property of hashes in general, only in this particular implementation of ruby, and I don't want to set them up with bogus expectations.

There's probably a better way to address this, and I think you're right that it involves talking about the "why" of it.

01-ruby-fundamentals/data-structures.md
```diff
@@ -1,211 +1,312 @@
...
+- The array will keep track of the order elements were added
 
 ```ruby
-# declaring an Array
-my_list = [ "Bulbasaur", "Ivysaur", "Venusaur", "Charmander", "Charmeleon", "Charizard", "Squirtle", "Wartortle" ]
+students = ['Ada', 'Katherine']
 
-# Accessing the first element
-puts my_list[0]
+puts "#{students}"
```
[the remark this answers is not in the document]
droberts-ada: For arrays and hashes, puts with interpolation will put it on one line with brackets, making it much more visually obvious (IMO) what's going on. Puts with no interpolation will split it across many lines with no brackets.

```ruby
myarr = [1, 2, 3]
puts myarr
# Output:
# 1
# 2
# 3
puts "#{myarr}"
# Output:
# [1, 2, 3]
```

01-ruby-fundamentals/data-structures.md
CheezItMan: https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas

droberts-ada: Good question. This is why we need a style guide.

droberts-ada: Rubocop says no by default, but you can make it OK for multiline literals
https://github.com/bbatsov/rubocop/issues/713

droberts-ada: Crap chris had already answered this when I hit send. I guess our conclusions are the same - I'll remove the trailing commas.

PR #371, opened 2018-02-08
Updating our Ruby Gems Section

CheezItMan:
## Description
While I liked students researching and presenting gems, we need to cut time here as we compress the curriculum.  

I've cut the gem section to walk students through 4 Gems, awesome_print, colorize, pry and faker.  We could cut any if needed.  I also have students practicing using the gems and the github documentation as a bit of practice in figuring out how to make a gem work.  This should cut a 3-hour section down to 1.


01-ruby-fundamentals/intro-to-ruby-gems.md
```diff
@@ -71,15 +116,86 @@ puts "This is blue text on red".blue.on_red.blink
 puts "This is uncolorized".blue.on_red.uncolorize
 ```
 
-## Activity: Divide & Conquer Ruby Gems
-In pods, research the usage of one of the following gems and create a small example showing at least one of the features of the gem.
+Relatively straight-forward, but neat!
```
kariabancroft: Let's not use this language, since it makes an assumption that isn't necessarily true for a student.

CheezItMan:
@tildeee Yeah, but I wanted to end it with an exploration of a Gem and figuring it out from the README, but Pry is hard to learn that way.  Like I said we could drop any of the gems and we do work with Pry later, but  OTOH I kinda feel it's nice to have Pry replace IRB and it's good that they know what it is, when we introduce it later for debugging.

PR #375, opened 2018-02-16
Cm/update inheritance

CheezItMan:
## Description
An update to the Inheritance lecture with better, I think, examples.  

## Questions for the Author
- [x ] Have you updated the weekly learning goals in the pedagogy resource

## Todos
Nope

## Feedback Requested
Feedback requested!


02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,131 @@
...
-- Use `super` to invoke the behavior of a _base class_ in a _subclass 
+- Use `super` to invoke the behavior of a _base class_ in a _subclass
 
 ## Object Inheritance: _is-a_
 _Inheritance_ is used to model the behaviors of one class after another. We would want to do this when two objects have a clear chain of ownership. Often the parent class, or _base class_, is more general while the child, or _subclass_, is more specific.
 
-For example, we could have a `Bird` class that describes the general behaviors of birds. This is a _base class_. We could create `Penguin` and `Hawk` classes that model more specific behaviors. A `Penguin` is very different than a `Hawk`, but they are both birds. It would make sense then, to declare them _subclasses_ of `Bird`. 
+For example, we could have a `LibraryItem` class that describes the general behaviors of items in a library. This is a _base class_ or _superclass_. We could create `Book` and `Video` classes that model more specific behaviors. A `Book` is very different than a `Video`, but they are both items in the library. It would make sense then, to declare them _subclasses_ of `LibraryItem`.
```
kariabancroft: I really like that you're referencing two different objects that could inherit this behavior. Could we include Video in the code sample and diagram as well? Just in the first set of things, then in the later more complex examples, we could say that we are just going to focus on Book and leave off Video for now.

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,131 @@
...
-tux.talk # => "I'm a bird. Kah."
+poodr = Book.new
+poodr.barcode #=> "12345"
 ```
 
-This is known as _implicit inheritance_ because when we put functionality in the _base class_ (`Bird`), then all _subclasses_ (`Penguin`) will automatically get those behaviors.
+This is known as _implicit inheritance_ because when we put functionality in the _base class_ (`LibraryItem`), then all _subclasses_ (`Book`) will automatically get those behaviors.
 
 ## Overriding Methods
 What happens when our classes have different functionality for the same idea? We can _override_ methods from _base classes_ in their _subclasses_.
```
kariabancroft: Could we add some more explanation about why we'd want the barcode to be differentiated from one object to another?

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,131 @@
...
 
 ```ruby
-class Penguin < Bird
-  def talk
-    "I'm a bird. Sqwaaa."
+class Book < LibraryItem
+  def barcode
+    return "BOOK12345"
```
kariabancroft: This could actually be written to use super in the exact way you do the audio book example. could you make the numbers distinct here to highlight how this is totally different than the barcode method in LibraryItem?

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,131 @@
...
 
-A way to describe this is by using the _is-a_ phrase: `Penguin` _is-a_ `Bird`. `Hawk` _is-a_ `Bird`. In Ruby, we use the `<` symbol to denote that the class on the left of the `<` symbol is inheriting behaviors from the class on the right side of the `<` symbol:
+A way to describe this is by using the _is-a_ phrase: a `Book` _is-a_ `LibraryItem`. `Video` _is-a_ `LibraryItem`. In Ruby, we use the `<` symbol to denote that the class on the left of the `<` symbol is inheriting behaviors from the class on the right side of the `<` symbol:
 
 ```ruby
-class Bird
-  def talk
-    "I'm a bird. Kah."
+class LibraryItem
+  def barcode
```
kariabancroft: I think i'm concerned because this seems like it should really be an instance variable to start out with. It might be more confusing that it isn't to start out.

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,138 @@
...
-class Bird
-  def talk
-    "I'm a bird. Kah."
+class Product
+  attr_reader :id, :name
+
+  def initialize
+    @id = 12345
+    @name = "Generic Product"
+  end
```
droberts-ada: Can you make `id` and `name` be parameters for the constructor? Having them hard-coded feels really goofy.

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,138 @@
...
+In the above example notice the `super` keyword in the 1st line of the `Book` class' `initialize` method.
+
+In initialize, `super` calls the parent, or superclass' constructor.  So `Product`'s `initialize` method is called and `@id` and `@name` are set to the given parameter.
+
+`super` must be the first line in a subclass' `initialize` method.  If `super` is called without an argument, it will use the parameters from the subclass' `initialize` method.
+
+![Super in initialize ](images/inheritance-super.png)
+
+**Exercise**  With your seatmates create an `Video` class.  A `Video` will have, in addition to an ID number and a name, a director and publisher.  It will also have a method which returns the format of the video (BlueRay, DVD, Streaming).  The `about` method should list the format of the video, the ID and the name. Then create an instance of the `Video` class.
+
```
droberts-ada: Why would format be a method here - seems like it should be a property. I also don't know that making it a method really demonstrates any of the learning goals.

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,138 @@
...
+audiobook.about # => "12345: Book-Audio"
+```
+
+In this case, we utilize the behavior that the _base class_ (`Book`) provides but augment it with behavior specific to the _subclass_ (`AudioBook`).
+
+![super in method calls](images/super.png)
+
+## Super & Initialize
+
+We've seen how methods by themselves work with methods, what about instance variables?  Instance variables are inherited as well, for example:
```
droberts-ada: The section heading here talks about super and initialize, but the text talks about instance variables. This feels confusing to me.

02-intermediate-ruby/object-inheritance.md
```diff
@@ -3,77 +3,138 @@
...
 What happens when our classes have different functionality for the same idea? We can _override_ methods from _base classes_ in their _subclasses_.
 
 ```ruby
-class Penguin < Bird
-  def talk
-    "I'm a bird. Sqwaaa."
+class Book < Product
+  def about
+    return "#{id}: Book"
   end
```
droberts-ada: The implementation of `Book#about` feels a little contrived to me. Is there a way we could provide a more intuitive example for these `Overriding Methods` and `Using super` sections?

Here is one idea that is pretty similar in complexity, but maybe a little more natural:

- Setup (previous section)
  - Each product knows what aisle it lives on
  - Product has a `find_in_store` method, that returns `"aisle #{@aisle_number}"`
- Overriding Methods
  - For a book, knowing the author's name is helpful in finding it on the shelf. We need to change the `find_in_store` method to give this information.
  - Now return `"aisle #{@aisle_number}, author #{author}"`
  - This also demonstrates that you can use instance variables from the parent in the child
- Using `super`
  - This isn't very DRY
  - Change it to `super + ", author #{author}"`

In order to make this work you'll need to move the bit on constructors up to before you talk about overriding methods. This is OK - it's a special case, and showing the pattern of "call `super` in the constructor" early is probably a good thing.

Let me know what you think.

droberts-ada:
Here is a question: Is this example going to seem silly or confusing given that we give this lesson in the middle of the grocery store project?

02-intermediate-ruby/object-inheritance.md
CheezItMan: I can, I just didn't want to jump into using `super` in the subclass' constructor initially.

CheezItMan:
@droberts-ada Yep, so I've switched it to properties, like in Redfin or Zillow.

PR #377, opened 2018-02-20
proposal for pair feedback models

tildeee:
## Description
We wanted a way that students could give/receive feedback:
"Another opportunity for the students to get real feedback on their project from someone who is not on the instructional team, have it documented, and have it presented in a meaningful way. Both code feedback and not-code feedback"

I propose the following system:
- giving the students this resource for giving/receiving feedback pro tips
  - i wrote a bunch of specifics for feedback (like be specific, etc), but also put a lot of quotes from other resources so it didn't look like i was making up stuff
- requiring the students to spend at least one 15-30 minute session for feedback
- the feedback template for the project (Scrabble) now asks for two things they discussed during their feedback session that they'd be willing to share

## Feedback Requested

Two things:

1. This document concretely says **"The pair/group needs to take 15-30 minutes to give and receive feedback. ... At Ada, we'll explicitly ask each pair/group to spend time giving/receiving feedback with each other at least once before project submission."** and I mostly want confirmation that that's a decent/accurate thing to say.

2. I made Scrabble PR comprehension questions have the following question: "What are two discussion points that you and your pair discussed when giving/receiving feedback from each other that you would be willing to share?" -- Is that fair?
https://github.com/Ada-C9/Scrabble/blob/master/.github/PULL_REQUEST_TEMPLATE

Thanks!

### DIFF
--- a/03-leadership-and-inclusion/pair-feedback-and-code-reviews.md
+++ b/03-leadership-and-inclusion/pair-feedback-and-code-reviews.md
@@ -1,0 +1,106 @@
+# Pair Feedback and Code Reviews
+
+When we pair program or program collaboratively professionally, we want to open as many opportunities to give and receive feedback as possible.
+
+As part of pairing or group projects, the pair/group needs to take 15-30 minutes to give and receive feedback. It is best to give and receive feedback after any activity or session of pairing concludes, while the session is still fresh in your memory. If you don't have the chance to give/receive feedback immediately after a pairing session, take notes while you remember.
+
+At Ada, we'll explicitly ask each pair/group to spend time giving/receiving feedback with each other at least once before project submission.
+
+Feedback can be about:
+- Code
+- The pairing experience
+
+Feedback should always be:
+- Actionable
+- Specific
+- Kind
+- Empathetic (for yourself and for others)
+
+Topics to cover are:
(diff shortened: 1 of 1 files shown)


03-leadership-and-inclusion/pair-feedback-and-code-reviews.md
```diff
@@ -0,0 +1,120 @@
...
+1. Be ready to learn and to teach
+1. Give and receive feedback when you're done
+
+[Tarso Aires, 10 Ways to Improve Your Pairing Experience](https://www.thoughtworks.com/insights/blog/10-ways-improve-your-pairing-experience)
+
+## How to Give a Code Review
+
+Code reviews are one of the best ways to provide feedback on someone's work. There are a lot of tools to assist with making code reviews, such as the review/comment system on pull requests on GitHub.
+
+It is not part of the curriculum at Ada to give code reviews. Even so, we want to prepare everyone to be able to contribute to giving feedback on someone else's code.
```
droberts-ada: Students don't CR each other, but we do CR them (kind of). Might be worth pointing out here.

kariabancroft:
[Peer Feedback](https://trello.com/c/zqKxaynh/600-peer-feedback)

PR #391, opened 2018-03-20
add strong params resource and update overall readme

kariabancroft:
## Description
Strong params!

## Questions for the Author
- [ ] Have you updated the weekly learning goals in the pedagogy resource - nope! 

## Todos
N/A

## Feedback Requested
Nothing specific

### DIFF
--- a/08-rails/forms-strong-params.md
+++ b/08-rails/forms-strong-params.md
@@ -1,0 +1,81 @@
+# Rails Forms: Strong Params
+
+## Learning Goals
+- Identify what strong parameters are and why Rails requires them
+- Extend Rails forms to use strong parameters
+
+### Getting Started
+
+Let's consider where we ended up most recently with our controller method that handled adding a new book, `create`. Here is the code we ended up with:
+```ruby
+# in app/controllers/books_controller.rb
+def create
+  @book = Book.new(author: params[:book][:author], title: params[:book][:title]) #instantiate a new book
+  if @book.save # save returns true if the database insert succeeds
+    redirect_to root_path # go to the index so we can see the book in the list
+  else # save failed :(
+    render :new # show the new book form view again
+  end
+end
+```
+
+This code is written with the assumption that our `params` hash contains data as follows:
+```ruby
+{ book:
+  {
+    author: "some author",
+    title: "some title"
+  }
+}
+```
+
+Notice that the parameters passed into `Book.new` are all listed out individually. What if we wanted to pass all of the data from the `params` directly without listing out each individual field? What happens when we try: `@book = Book.new(params[:book)])`?
+
+What happened? Most likely, you've gotten a **ForbiddenAttributeError**.
+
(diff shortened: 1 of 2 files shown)


08-rails/forms-strong-params.md
```diff
@@ -0,0 +1,81 @@
...
+### Strong Parameters
+Rails has introduced the concept of **strong parameters** to use in our controllers to prevent these errors. Strong params will require us to give explicit access to a set of attributes that come through `params` that we'll then be able to pass to our model methods. We set up strong params once for our controller, and then we will no longer need to specify each field individually (in the way we had to above).
+
+### Update the Controller
+Now that we have an idea of _why_ we need strong params, let's see how to implement these in our controller.
+
+1. Create a `private` method section towards the bottom of our controller.
+2. Create a new method that we'll use to return the _permitted_ parameters.
+    ```ruby
+    private
```
droberts-ada: As long as everything is preceded by 4 spaces, you can make all the numbers `1.` and it will do the numbering automatically.

08-rails/forms-strong-params.md
```diff
@@ -0,0 +1,81 @@
...
+    ```
+3. We'll use two special controller methods here which will allow us to define exactly which parameters (coming from the form) we want to allow. Those methods are `require` and `permit`.
+
+  We use `require` to identify the highest-level param which contains all others and that we cannot live without. We then use `permit` to list the params within that overall one which we want to allow.
+    ```ruby
+    private
+
+    def book_params
+      params.require(:book).permit(:author, :title)
+    end
```
droberts-ada: Could you use an explicit return here? I think that makes it more obvious what this code is doing.

08-rails/forms-strong-params.md
```diff
@@ -0,0 +1,81 @@
...
+      ...
+      # this code is the same
+    end
+    ```
+5. Run through the form submission again to ensure that we are no longer getting the `ForbiddenAttributeError`.
+
+A few things to note here:
+1. The new `book_params` method is using the `params` variable which the controller makes available to us (even though we're not passing it in as a parameter)
+2. We chain the method calls of `require` and `permit`. This works because the `:author` and `:title` fields are located **within** the `:book` param.
+
```
droberts-ada: Again, should be able to use `1.` for both of these and let the numbers happen automatically.

08-rails/forms-strong-params.md
```diff
@@ -0,0 +1,81 @@
...
+  else # save failed :(
+    render :new # show the new book form view again
+  end
+end
+```
+
+This code is written with the assumption that our `params` hash contains data as follows:
+```ruby
+{ book:
+  {
```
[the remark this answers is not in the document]
kariabancroft: I wanted to show it like a hash so if they haven't seen params yet they can visualize how the data is structured

kariabancroft:
@droberts-ada I responded to your comments

08-rails/forms-strong-params.md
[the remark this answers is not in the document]
droberts-ada: Shouldn't `book:` be on its own line though?

PR #394, opened 2018-03-21
Cm/erd activity

CheezItMan:
## Description
An ERD Practice Activity to get'em to draw out a slimmed down ERD.   A sample solution is [here](https://www.lucidchart.com/invitations/accept/1c19a336-5d12-4e9c-ab62-4d32efbcedff)

## Questions by the Author
- [ ] Is this too small a problem?

## Feedback Requested

### DIFF
--- a/08-rails/exercises/erd-practice-activity.md
+++ b/08-rails/exercises/erd-practice-activity.md
@@ -1,0 +1,21 @@
+# ERD Practice Activity
+
+## Scenario
+
+We are setting up an application to allow people to share their homes to tourists online like AirBnb.  So our initial user stories would be:
+
+- As a user I can add a home to be listed for rentals, along with a price, so I can rent the home out and make money.
+- As a user I can choose when a home of mine is available for rentals so I can pick when I can have guests.
+- As a user I can can select a home to rent for a vacation on a specific set of dates.
+- As a user I can see all the money I have spent.
+- As a user with rentals I can see all the money I have made from clients renting my homes.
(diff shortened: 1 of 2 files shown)


08-rails/exercises/erd-practice-activity.md
```diff
@@ -0,0 +1,21 @@
+# ERD Practice Activity
+
+## Scenario
+
+We are setting up an application to allow people to share their homes to tourists online like AirBnb.  So our initial user stories would be:
+
+- As a user I can add a home to be listed for rentals, along with a price, so I can rent the home out and make money.
```
kariabancroft: I think "home or property" would be a good descriptor here. "listed for rentals" is not quite right, I think we want "listed for rent" here and on the line below. I don't know if we need the last portion

08-rails/exercises/erd-practice-activity.md
```diff
@@ -0,0 +1,21 @@
...
+- As a user I can choose when a home of mine is available for rentals so I can pick when I can have guests.
+- As a user I can can select a home to rent for a vacation on a specific set of dates.
+- As a user I can see all the money I have spent.
```
kariabancroft: I'd group the top 3 bullets about setting up a rental together and then the bottom three together about reserving a rental

CheezItMan:
Updated as per Kari's comments,

Part of the point is that an individual user can be both a host and a guest at different times.

PR #399, opened 2018-03-29
Split resources on nested routes and many-to-many relations

droberts-ada:
## Description
Previously these were one lesson, but this didn't make much sense to me. It was introducing multiple concepts at the same time, and it didn't seem like the two problems matched up well.

Instead, I've used the existing books-authors relation to talk about nested routes. Since the example is so different this involved almost a total re-write, but I think this is OK.

I did use the books-genres relation to discuss many-to-many relations. It goes through setting up the relation in much more detail, and has a substantial piece on building UI elements that was absent before.

## Questions for the Author

## Todos
- [x] Lesson on model logic

## Feedback Requested
If someone were to run through the examples and verify they work, I would appreciate it.

One thing I didn't mention in here is `has_many through:`. We've got an opportunity to demo it with the authors-genres relation (an author `has_many :genres, through: :books`), but the lesson is pretty long as-is, they already don't get an opportunity to practice on the project, and I'm hesitant to pile too much more in there. If you think it would be useful I can add it pretty easy.


droberts-ada:
[Nested Routes and Many-to-Many](https://trello.com/c/yNIichpo/748-nested-routes-and-many-to-many)

08-rails/exercises/active-record-relationships-exercise.md
```diff
@@ -0,0 +1,151 @@
...
+    # app/controllers/books_controller.rb
+    class BooksController < ApplicationController
+      # ... all the actions ...
+      private
+      def book_params
+        # Fill in other field names as appropriate
+        return params.require(:book).permit(:title, :author_id)
+      end
+    end
+    ```
```
kariabancroft: Should this have some sort of conclusion?

droberts-ada:
[Rails week 2 curriculum updates](https://trello.com/c/dgOxSfdc/469-rails-week-2-curriculum-updates)

08-rails/exercises/active-record-relationships-exercise.md
droberts-ada: Probably. This exercise already existed, I just moved it into the `exercises` folder.

PR #402, opened 2018-04-06
Update resource on model testing

droberts-ada:
## Description
The idea is to make a "typical" instance of the model in a `before` block, then each test `nil`s something out and verifies it isn't valid.

## Feedback Requested
Nothing special


droberts-ada:
[Add guidelines around Rails Model Testing](https://trello.com/c/LtUhYLEQ/474-add-guidelines-around-rails-model-testing)

PR #404, opened 2018-04-23
Resource on debugging based on user reports

Hamled:
## Description
New resource explaining how to take user reports of erroneous behavior and turn them into repro steps for a developer to use when fixing bugs.

This resource is intended to be used as part of the Rails Debugging activity, since the bugs included in that activity are presented to the student as user reports.

## Feedback Requested
Let me know if there are any more details about how this process works that I should include!

### DIFF
--- a/00-programming-fundamentals/debugging-user-reports.md
+++ b/00-programming-fundamentals/debugging-user-reports.md
@@ -1,0 +1,43 @@
+# Debugging from User Reports
+Debugging any application usually takes one of two forms: either you encounter the bug yourself while developing the application, or a user encounters the bug and files a report. You've already seen the first form of debugging quite a lot here at Ada because it's a natural, necessary part of building software.
+
+The second form of debugging is something that you will likely begin to experience during your internship, and it has some unique aspects to it that inform the approach you'll need to take when debugging. Specifically, it is necessary to translate from the user's report into a set of "reproduction steps" (also called a "repro") which can be followed to trigger the bug.
+
+In some organizations this work is the purview of dedicated customer support and/or QA team members, but in smaller companies the responsibility often falls on developers as well. In either case it is important to get this experience to be able to work effectively with the groups that support the live/production deployment of the system or application.
+
+## Creating Repro Steps
+We need to create these repro steps because users of our application generally do not have any understanding of how it actually works. They only see the user interface and so their explanation of what went wrong is usually framed in terms of how things are presented in the UI.
+
+For example, they might make a purchase from your store and see that it does not show up in their purchases list. They might then report this to you as a bug with the purchase list. But upon investigation you might find that the issue was actually due to the payment gateway system returning an error code that your application did not expect, resulting in the purchase being dropped from the database.
+
+Once you have a set of repro steps for a specific bug you can follow them in your local development environment and see the full error message and details (which are generally hidden from users in production deployments). At this point you can follow the same debugging process as when you encounter an error yourself.
+
+## Example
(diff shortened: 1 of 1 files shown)


Hamled:
[Rails debugging activity](https://trello.com/c/G9MuVbzX/104-rails-debugging-activity)

00-programming-fundamentals/debugging-user-reports.md
```diff
@@ -0,0 +1,43 @@
...
+1. Sign up for a new account using the standard sign up procedure.
+1. Use the product search bar to search for "silver tails dog infrared massager"
+1. Click on the result labeled "Silver Tails Dog Infrared Massager with 2-Heated Heads."
+1. Click on the "Add to Cart" button.
+1. Press the browser's back button.
+1. Click on the "Add to Cart" button again.
+1. Observe that the quantity for the above item is listed as 2.
+1. Open the quantity drop-down for the item and select "1".
+1. Click the "Update item" button next to the drop-down.
+1. Observe that the item "Silver Tails Dog Infrared Massager with 2-Heated Heads." is no longer listed in the cart.
```
droberts-ada: Maybe have explicit "expected behavior" and "observed behavior" pieces here? In practice I've found explicitly listing off what we expect to have happen is helpful. It lets you know when the bug is really fixed, and potentially gives you a test case.

00-programming-fundamentals/debugging-user-reports.md
```diff
@@ -0,0 +1,43 @@
...
+
+  Sometimes if a developer receives an issue ticket (how bugs and feature work are tracked) with some lenience or variation in how the repro steps could be followed they will try it one way and not trigger the bug, and then they will mark the issue as resolved because it's not a real bug.
+* The repro steps should involve the original report's details to the greatest extent possible. In these repro steps we continued to use the dog massager item, even though this bug might affect other items.
+
+  While you can sometimes generalize the repro steps, it's not always obvious when that is possible or not. In those cases err on the side of caution and re-use the details from the initial report -- they might be critical to diagnosing the underlying bug.
+* The repro steps should conclude with an observation. Generally the final part of any set of repro steps is to observe the unexpected behavior, this makes it clear specifically how the bug is manifesting (at a user level) in the final context of the repro steps.
```
droberts-ada: Adding a vocab list here would be good. Things like:

```
### Vocab
- **User report:** A description of a bug found by a user. Often described in terms of the UI, since the user doesn't have any insight into how your app actually works.
- **Repro:** A set of steps to consistently reproduce a bug.
- **Observation:** The last step of a repro. Includes a description of what was expected and what actually happened.
```

droberts-ada:
It also might be worthwhile to add some explicit learning goals.

Hamled:
My goal was definitely for it to feel relevant to students when they're working on the activity. The activity's instructions for the example bug fix are pretty explicit about creating repro steps.

I was not thinking that this resource should be presented separately from the activity, though. I only moved it into a separate textbook resource because the README for the debugging activity is already very long.

00-programming-fundamentals/debugging-user-reports.md
Hamled: Yeah I had thought about this while writing it, and then shied away from it I guess because usually the "expected vs observed" stuff is in a separate section of the issue ticket's template.

However, since the students don't yet have that context it makes sense to discuss it explicitly. I will add that in.

Hamled:
@droberts-ada @tildeee PR has been updated to reflect your review feedback.

PR #405, opened 2018-04-24
Cm/js testing to jest

CheezItMan:
## Description
Last week I played a bit with Jest and took a stab at updating the testing resource.  I'd like to do some basic testing with vanilla JS and then once we have done some React reintroduce testing React components, so students have some familiarity with it once they start React.   I've also made a branch of the [js-scrabble](https://github.com/AdaGold/js-scrabble/tree/cm/update-to-jest) project updated to use Jest with the last few tests stubbed out for students to complete.  

It's very similar to minitest, and they will see virtually identical testing in Jasmine in the Exercisms.  In looking at this I wanted to frontload as much as possible to minimize new content with React both to minimize new material in the React section and to provide a resource on testing since they will see them in the exercisms and Scrabble already.  

## Questions 
- Is this too much too early?  

## Questions for the Author
- [ ] Have you updated the weekly learning goals in the pedagogy resource
- [x] Do we want to do testing in vanilla JS prior to React.  

## Todos
- [ ]  Still need to finish a React testing section.

## Feedback Requested


CheezItMan:
[JS Testing to Jest](https://trello.com/c/Q9Bl3DgJ/800-js-testing-to-jest)

CheezItMan:
[JS Testing to Jest](https://trello.com/c/Q9Bl3DgJ/800-js-testing-to-jest)

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
 
 ### Testing Strategy
 
-We'll be using a test suite called [Jasmine](https://jasmine.github.io/index.html) (maintained by Pivotal Labs, part of the Ada network). Jasmine is available as an npm package, and it is included in our standard Ada webpack setup.
+We'll be using a test suite called [Jest](https://facebook.github.io/jest/docs/en/getting-started.html) (maintained by Facebook). Jest is available as an npm package and comes built-in with `create-react-app`.
```
kariabancroft: I don't think they'll have any context to `create-react-app` yet, so we should leave this out

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
+We'll start testing using the second technique. It will be very reminiscent of model testing in Rails, and in fact we will focus our JavaScript testing on business logic methods. The testing techniques we learn here will also transfer directly to any command-line JavaScript programs you write, including something like a Node-Express server.
 
 As an aside, Selenium is a *really* cool technology, we just don't have the time to do it justice. Maybe it would make a good piece of a capstone project...?
 
 ### Setup
-We will utilize the same basic application structure that we have used for our Backbone applications. In fact, the project we'll be working with is a complete Backbone SPA. To let us focus on the process of writing tests with Jasmine rather than implementing some bit of complex logic, we'll be using a Backbonified version of our good old Scrabble game.
+We will start writing tests with a sample whiteboarding problem.  
 
-1. Clone this repo: `https://github.com/AdaGold/scrabble-jasmine-baseline`
+1. Clone this repo: `https://github.com/AdaGold/pangram`
```
kariabancroft: Make this a link please

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
 1. `npm install`
```
kariabancroft: These steps seem to be duplicating and different than the steps in the pangram readme. Recommend having one single set of instructions that include all of the steps we want them to run, either here or in the repo itself.

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
+Examine the README.  
 
-Now look through the code, and confirm or rebuke your suspicions. Pay close attention to the `spec` directory, which is where the tests live - we'll dive into it a little later. To run the tests, type `npm test` at the console.
+**Question**: In testing a Pangram, what edge-cases would you have?  
 
-Similar to Rails, the `spec` directory has the same structure as the `src` directory. Each file `src/dir/name.js` may have a corresponding `spec/dir/name.spec.js`.
+Now look through the code in the `pangram.spec.js` file, and see if the tests described match your predictions. To run the tests, type `npm test` at the console.
 
-Because we're running tests via `npm`, our tests have access to all our WebPack infrastructure, and `import` and `export` statements will work about as you'd expect.
+Similar to Rails, the `spec` directory has the same structure as the `src` directory. Each file `src/dir/name.js` may have a corresponding `spec/dir/name.spec.js`.
```
kariabancroft: There is no `src` directory, so if you want to see this parallel this should be added to the file structure in pangram

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
+Because Jest is a behavior-driven testing language, testing using Jest for JavaScript is fairly similar to what we saw using Minitest for Ruby and Rails. As always, there are some specific differences.
 
 Each test should have these components to describe the test behavior.
 
 #### `describe` Blocks
-Describe blocks should create test groupings based on _objects_ and _functions_.
+Describe blocks should create test groupings based on _objects_ and _functions_.  They are optional, but provide some readability and organization.  Like Minitest you can also nest `describe` blocks.
 
-We should use one `describe` block for our overall `Word` model, and another nested one for each method or set of behavior we want to test. `Word` has one for the constructor, another for validations and a third for the `score()` method.
+We will use one `describe` block for our overall `Pangram` class, and another nested one for the `isPangram` method.
```
kariabancroft: Overall - I struggle with using this example since I think that having a method inside this class for this is really quite unnecessary. I can't really justify to the students why they would create an instance of a class with a string, just to call the method on this, rather than just create a method that takes a single string parameter. Do you feel like this is better than one of the Scrabble examples?

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
 ```
 
+Similar to what you have in Minitest you can use `it`, in place of `test` if `it` (hic) makes you feel more comfortable, but the Jest documentation uses `test`, and so Ada's examples will as well.
+
 #### `expect` Statements
 
 Expectations should be the "meat and potatoes" of your tests, inside of your `it` blocks. Each test has at least one `expect` statement to ensure the behavior is as expected.
 
-The syntax of `describe` and `it` is pretty similar to Minitest (at least as similar as Ruby and JS can be), but `expect` is somewhat different. Note that there are actually two functions being called, the `expect` and the matcher (`toEqual` in this case), each taking one argument.
+The syntax of `describe` and `test` is pretty similar to Minitest (at least as similar as Ruby and JS can be), but `expect` is somewhat different. Note that there are actually two functions being called, the `expect` and the matcher (`toBeDefined` in this case).
```
kariabancroft: Instead of this sentence with the "Note", can you explicitly state how the `expect` statement is constructed? (I know that you're pulling from the notes as they previously existed, but I feel like this could use a more explicit explanation)

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
+    expect(Pangram).toBeDefined();
+  });
+
+  describe('isPangram()', () => {
+
   });
 });
 ```
 
 And that's all a test is. Go ahead and add another one to the spec file, but this time make sure it will fail (e.g. `expect(false).toEqual(true);`). Then re-run the tests, just to see what a failure looks like. Fix the test and run it again. Not too different from Minitest, right?
```
kariabancroft: "add another `expect` statement" or "add another assertion"

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
+Tests can be skipped by changing `test(...` to `test.skip(...`.  Go ahead and change
+
+```javascript
+    test('isPangram() is defined', () => {
+```
+
+to:
+
+```javascript
+test.skip('isPangram() is defined', () => {
```
kariabancroft: Mismatched indentation to example above

10-JavaScript/testing.md
```diff
@@ -23,96 +22,119 @@ In short, most interesting JavaScript code is full of *external dependencies*. M
...
+```
+
+If you choose to use `it` blocks instead of `test`, you can skip a test with `xit` in place of `it`.
```
kariabancroft: I think we should just leave this out with the assumption that we're going to focus on using `test` over `it`

10-JavaScript/testing.md
```diff
@@ -123,24 +145,18 @@ Just like in testing with Ruby and Rails, Jasmine has a number of **matchers** t
...
+[https://facebook.github.io/jest/docs/en/expect.html](https://facebook.github.io/jest/docs/en/expect.html)
 
-Now use a test-driven development workflow to implement the `Word.score()` method. Remember: the TDD cycle: pseudocode-red-green-refactor!
+## Exercise With A Seatsquad Partner
 
-Feel free to borrow production (non-test) code from either of the previous times we've implemented scrabble - the learning objective is to practice Jasmine testing syntax, not to build the Scrabble logic for the third time.
+Read through the couple of implemented tests in `pangram.spec.js` and the stubbed out tests.
 
-Once we've spent some time here, we'll come back as a class and compare results.
+Now use a test-driven development workflow to implement the `Pangram.isPangram()` method and complete the stubbed-out tests. Remember: the TDD cycle: pseudocode-red-green-refactor!
```
kariabancroft: Remove the first : after Remember

10-JavaScript/testing.md
tildeee: Would it make more sense for you if we refactored the `Pangram` class to `TextBlock` or `Snippet` or something else that might have a `isPangram` method and implies there are other purposes for it?

CheezItMan:
Both the AdaGold repo & the textbook materials are updated.

tildeee:
My two comments are:

1. It's unfortunate that the project `TextBlock` is defined very closely to when we talk about test blocks :P lol
1. I don't think this activity fits in an hour -- do we want to adjust and extend the activity to 1.5 hours?

CheezItMan:
I think it'll actually fit in just over an hour as introducing the formatting would only take 30 minutes or so, and 30 minutes to write a few tests for Pangram won't take them too long once they get over callbacks with arrow functions.

PR #407, opened 2018-05-07
Update the rails template to work with APIs

droberts-ada:
Previously it would break when trying to add code to non-existent JS and CSS files. Now it does not do that.

### DIFF
--- a/09-intermediate-rails/reference/.rails-template.rb
+++ b/09-intermediate-rails/reference/.rails-template.rb
@@ -1,24 +1,32 @@
-# Make $(document).ready work as expected, despite turbolinks weirdness
-gem 'jquery-turbolinks'
+puts "Executing the Rails template"
+API_MODE = ARGV.include? '--api'
+puts "API mode: #{API_MODE}"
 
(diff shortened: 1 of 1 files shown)


tildeee:
I didn't give this a thorough review but I just wanted to say that I saw it, it looks like it works, and it's a pretty substantial change? Dang, weird

Also, do we need to make this a different rails template? For instance, C9 will now replace their first version (potentially second) version of our rails template-- for C10 do we want a similar journey, or will we give them this from the start?

CheezItMan:
It's a good change, and one that will prevent things like Foundation, and jQuery gems from being installed.  Useful, but for 1 week of lessons... I'm not sure it's worth the time.  

We could introduce this at the beginning of C10, but that would make it more complicated to walk through for students.

CheezItMan:
Also I tested it and it works with Ada Pets

droberts-ada:
@tildeee It's mostly whitespace changes. There are a couple places where I moved things around or added a variable name, but I don't think I've added or removed anything.

`git diff --ignore-all-space` shows it clean.

PR #413, opened 2018-05-16
Split out props and state

kariabancroft:
## Description
Split these two resources. Ensure that the state resource stands alone with the assumption we've already learned props.

## Questions for the Author
- None

## Todos
N/A

## Feedback Requested
N/A


React/props.md
```diff
@@ -0,0 +1,59 @@
+# Passing data using `props`
+
+## Learning Goals
+- Examine how React allow components to manage data
```
Hamled: allow -> allows

Hamled: allow -> allows

Also, perhaps "use data" or "receive data" would be better than "manage data", since props doesn't allow the component to modify the data at all.

React/props.md
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div>
+        The data from props is: { this.props.vaName }
+      </div>
+    );
+  }
+}
+```
+
+**More Complex**
```
Hamled: Can we give an exercise prompt here for students to practice building a component which receives and uses a `prop` variable from another component?

Something simple like:

Create a `Link` component which uses two `prop` variables, `text` and `url`, to create an `a` tag which links to the given URL. Have another component (such as `App`) render multiple `Link` components.

React/state.md
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+```
+
+
+Next, we'll need to add the event handler function that our code will call when the button is clicked.  
+
+```javascript
+// between constructor and render
+onToggleButton = () => {
```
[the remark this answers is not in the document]
kariabancroft: Yes - see my question a few days ago in Slack about binding to `this` for event handlers

PR #416, opened 2018-05-17
Update AJAX lesson to use axios

droberts-ada:
## Description
Had to totally rewrite all the code, so took it as an opportunity to update the lesson structure as well. I've added code that lives under `resources/axios`. The HTML there is what we'll be using for both lessons, and the final version of the JavaScript lives there as well.

I've tried to focus on the end-to-end process of making an API call, including status updates, printing data to the screen, error handling, etc.

Pets repo has been updated.

## Questions for the Author

## Todos
None

## Feedback Requested
- Run through the code and make sure it's OK / all works
- Does the physical structure where stuff lives make sense? I thought about putting the code in a separate repo, but since it's only 2 files (3 w/ CSS) I didn't think it was worth it.


10-JavaScript/ajax-intro.md
```diff
@@ -1,217 +1,164 @@
...
 
+The `fetch` function was added to the core JavaScript language in the last few years as a replacement for XHR. `fetch` modern syntax and functionality without loading an external library. However, `fetch` provides a much lower-level view of HTTP requests than we need, and ends up being somewhat complex to use.
 
-## Let's Make an AJAX call!
+**In this course, we will be making AJAX requests using the axios library.**
 
-Note that in order for AJAX to work, we must make requests to a running server. It can be local or online, however we cannot make requests to a file on our computer.
+## AJAX with axios
 
+All of the JavaScript we write in this lesson and the next will reference the HTML and CSS files in the [reference/axios](reference/axios) folder. If you want to follow along, you should go ahead and copy that code now.
```
CheezItMan: I kinda feel like it might be easier if we had a repo or codepen, so students could reference any sample-code we write in class.  

This will definitely work however.

10-JavaScript/ajax-intro.md
```diff
@@ -1,217 +1,164 @@
...
 
-This is how to do it using vanilla JavaScript:
+With that HTML in mind, read through the following code and see if you can figure out what's going on. Specific questions to answer:
 
-```html
-<button id="load"> Fetch! </button>
-<div id="pets"></div>
-```
+- Program Structure:
+  - When does our code make the GET request? What would we need to change if we wanted it to happen as soon as the page loads?
```
CheezItMan: I like this section, but maybe this is a good place to do some think-pair-sharing.  We could challenge students to answer the questions below, and find a way to trigger the errors.

10-JavaScript/ajax-post.md
```diff
@@ -1,99 +1,185 @@
...
-Just like we used AJAX to send GET requests, we can also use it to send POST requests. Based on our knowledge of POST requests, what would you say is the main thing we would add to a POST request that we did not include in a GET request? YES! Data!
+## Introduction
+
+Just like we used axios and AJAX to send GET requests, we can also use it to send POST requests.
+
+**Question:** Thinking back to our previous work with APIs in Ruby, what are the big differences between POST and GET?
+
+## Making a POST request
+
+Let's get right to it! We will make a POST request to add a pet to the list. Read through the [API documentation on GitHub](https://github.com/AdaGold/pets_api/blob/master/README.md), and answer the following questions:
```
CheezItMan: Getting them used to reading an API doc.  👍

10-JavaScript/ajax-post.md
```diff
@@ -1,99 +1,185 @@
...
-  console.log('success!');
+// Inside $(document).ready() in index.js
+$('#pet-form').submit(createPet);
+```
+
+Next, we need to build the `createPet` method. We'll start out with hard-coded data, and later on we'll expand our code to read values from a form.
+
+```javascript
+// Above $(document).ready() in index.js
+const createPet = () {
```
CheezItMan: Missing an `=>`

10-JavaScript/ajax-post.md
```diff
@@ -1,99 +1,185 @@
...
+  // Later we'll read these values from a form
+  const petData = {
+    name: 'Socks',
+    age: 7,
+    owner: 'Bill'
+  };
+
+  reportStatus('Sending pet data...');
+
```
CheezItMan: You don't seem to be disabling the form submit so it refreshes the page.  i.e. `event.preventDefault()`.

That's not a bad thing, as showing them **why** that doesn't work is a good idea.

10-JavaScript/ajax-post.md
```diff
@@ -1,99 +1,185 @@
...
+    - Logging to the console as output
+1. Put together a user experience
+    - Read a form to get data
+    - Report progress and errors politely to the user via HTML
 
 ## Additional Resources
-- [jQuery AJAX POST Documentation](https://api.jquery.com/jquery.post/)
-- [jQuery AJAX Documentation](http://api.jquery.com/jquery.ajax/)
+- [axios documentation](https://github.com/axios/axios)
+- See resources from previous lesson
```
CheezItMan: Link?

Maybe also linking to jQuery docs for reference on the challenge above

droberts-ada:
[JS Week 2 axios updates](https://trello.com/c/CtZGlwll/835-js-week-2-axios-updates)

10-JavaScript/ajax-intro.md
droberts-ada: I think that is a good idea! My instinct was not to add classroom management directives into the lecture notes, but if you think it would be helpful I can.

10-JavaScript/ajax-post.md
droberts-ada: Whoops, definitely a typo!

PR #421, opened 2018-05-24
Cm/react forms

CheezItMan:
## Description
1st pass at forms in React.

## Questions for the Author

## Feedback Requested


React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+        </form>
+      </div>
+    );
+  }
+}
+
+export default NewStudentForm;
+```
+
+Now we have a React component to render the form.  It functions just like the HTML form, but can be imported into other React containers.  We term the name and email input fields _Uncontrolled components_ because the DOM is managing the values in the fields.  However now that we have them in a React component we can modify the form controls to have React monitor the input fields and handling the form submission.  To do so we will turn the inputs for name and email into _controlled components_.
```
kariabancroft: imported into React containers or components?

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+
+```javascript
+import React, { Component } from 'react';
+import  './new_student_form.css';
+
+class NewStudentForm extends Component {
+  constructor(props) {
+    super(props);
+}
+render() {
```
kariabancroft: Looks like indentation is a bit off in this ex

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+When a user enters text in the form controls the DOM stores the values of the form fields.  However we can do a lot more combining it with React.
+
+First we will build a React component to manage creating new students.
+
+```javascript
+import React, { Component } from 'react';
+import  './new_student_form.css';
+
+class NewStudentForm extends Component {
+  constructor(props) {
```
kariabancroft: I err on not passing `props` into the constructor unless it is a functional stateless component

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+    super(props);
+}
+render() {
+    return (
+      <div>
+        <form className="new-student-form">
+          <div>
+            <label htmlFor="name">Name:</label>
+            <input
+              name="name" />
```
kariabancroft: I'd put this on the line above for consistency

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+  this.state = {
+    name: '',
+    email: '',
+  }
+}
+```
+
+The `NewStudentForm` component will track the `name` and `email` as part of it's state.
+
+We can then manage the input components by setting the value of the form component to match the `NewStudentForm`'s state and adding an event handler to respond to changes in the input value.
```
kariabancroft: I don't think we refer to the elements in JSX as components. I think the input and the form are elements or fields not components.

kariabancroft: I think the confusion might be from the fact that this `NewStudentForm` component is becoming a controlled component overall - but I don't think we refer to each individual field using the same vocabulary.

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+So with our form we can track input into the fields and provide real-time validation feedback.  Next we want to actually **do** something with the data, but our `NewStudentForm` component doesn't, and shouldn't need to care what happens with the submitted data.
+
+**Question**:  Looking at our Ada student list app to date, what component would be managing the student data?
+
+## Handling Submissions
+
+Now we want to handle when the user submits the form.  We can add a function as an event handler.
+
+```javascript
+// new_student_form.js
```
kariabancroft: The component naming convention will be camel case (i messed this up in some of my notes)

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+...
+<form className="new-student-form" onSubmit={this.handleFormSubmit}>
+...
+```
+
+So we have a way to detect submit events on the form, but no way to get data to the rest of the application.  To solve this we will modify the `StudentCollection` component to pass a function as a prop to the `NewStudentForm` so it can pass the student data back.
+
+![passing a callback through props](./images/passing-callback.png)
```
kariabancroft: Can you put as an HTML comment the location where this image was constructed in case it needs editing later?

kariabancroft: This is a good diagram. Can we make it a bit bigger and make State lowercase?

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+
+First adding a callback function to `StudentCollection` and passing that function to `NewStudentForm` as a prop.
```
kariabancroft: First, let's add...

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+      <h3>Students</h3>
+      { studentComponents }
+      <NewStudentForm addStudent={this.addStudent} />
+    </div>
+  )
+}
+...
+```
+
+Then we can update the `handleFormSubmit` function.
```
kariabancroft: ... to call the function that came in from props.

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+// new_student_form.js
+...
+handleFormSubmit = (event) => {
+  event.preventDefault();
+  const newStudent = {
+    name: this.state.name,
+    email: this.state.email,
+  };
+
+  this.setState({
```
kariabancroft: A good question for them is why they'd need/want this

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+
+  this.props.addStudent(student);
+}
+```
+
+Now if we test the app we should now be able to add students to the list.  However there are some issues.
+1.  We can add students with no names
+2.  We can add students with invalid email addresses
+
+**Exercise**  Update the app to prevent form submission if the name is blank, or the email field is invalid.
```
kariabancroft: Maybe some guidance on which component they might want to think about adding this logic to

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
+
+## Final Note: Prop Validation
```
kariabancroft: I think prop validation is super valuable but it seems like maybe it should be a whole separate discussion. This very brief mention of it might bring up more questions than answers.

React/forms.md
kariabancroft: After going through the whole set of notes and then returning back to this discussion of controlled vs uncontrolled components, i'm not sure that we should present it right at the beginning. Could we demonstrate the pattern and then at the end give a name to it instead?

React/forms.md
```diff
@@ -2,24 +2,280 @@
 
 ## Learning Goals
 - Implement a form in a component
-- Handle the event of a form submission in a component
 - Dynamically provide user feedback as they complete a form
```
kariabancroft: Maybe clarify this to refer to validations?

React/forms.md
```diff
@@ -2,24 +2,280 @@
...
 | Term     | Definition     |
 | :------------- | :------------- |
-| Item One       | Item Two       |
+| **Controlled Components**       | Components whos state is managed by a React component.       |
+| **Uncontrolled Components**   |  Components who have their state stored in the DOM. |
+
 
 ## Key Takeaway
-Description of one or two sentences with key takeaway.
+Forms in a React application can either be controlled or uncontrolled.  Controlled components provide a number of capabilities beyound regular HTML forms.  Uncontrolled components can be easier to migrate from a non-react application.  Forms can be used to collect data and pass it back to containers via callback functions in the component's props.
```
kariabancroft: I'm not sure that the difference between the two is what they'll take away from this lecture.

CheezItMan:
@kariabancroft Updated

React/forms.md
```diff
@@ -2,24 +2,266 @@
...
+First we will build a React component to manage creating new students.
+
+```javascript
+import React, { Component } from 'react';
+import  './new_student_form.css';
+
+class NewStudentForm extends Component {
+  constructor() {
+    super();
+  }
```
tildeee: Do we need to define a constructor in this if all it does is call `super()`? This is an earnest question; I don't know React well enough

React/forms.md
```diff
@@ -2,24 +2,266 @@
...
+}
+```
+
+And give the user feedback on validation with:
+
+```javascript
+<input
+  onChange={ this.handleEmailChange }
+  value={ this.state.email }
+  className={ this.emailValid() ? "valid": "invalid" }
```
tildeee: Is this code in the `render()` function?-- we said in the material that you can't do conditional logic in render, so I wanna acknowledge that you can do these one-line ternaries because it's just JS that's evaluated inline

tildeee: I don't personally care either way, just since we said in the React Hello World Exercise (in Important Notes), someone will ask "didn't we say we can't do that?" so we should be able to say "yes we can" and it might help to have a note to talk about it in the lecture maybe?

https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/React/react-hello-world.md#what-is-jsx

CheezItMan: ok I added a note

PR #424, opened 2018-05-27
Updates to flow from state, events and student app

kariabancroft:
## Description
Use the state and events lectures to leverage a smaller more basic example of setting up state and impacting state with an event. All within a single component. 

The creating components part 2 lecture then takes this a steps further in combination with props and manages state within a parent component and then propagates the event from a child to a parent component. 

## Questions for the Author
N/A

## Todos
I might still want a diagram in part 2 of components part 2 to designate the flow, but this will also be fairly straightforward so i'm not adding it at this moment in favor of time. 

## Feedback Requested
Is creating components part 2 too complex?

### DIFF
--- a/React/creating-components-2.md
+++ b/React/creating-components-2.md
@@ -2,81 +2,73 @@
 
 ## Learning Goals
 - Continue nesting components within one another
-- Continue practice component creation and JSX syntax
+- Continue to practice component creation and JSX syntax
+- Use `state` within our components
+- Act on events within our components
+- Pass event handler functions as callbacks within `props`
 
-Next we are going to extend our application to support multiple students. Additionally, we are going to support components dynamically loading data.
+### Overview
+In this discussion, we are going to go back to our students application and add some new features. First, we're going to refactor our code a bit to set ourselves up to use some of the new concepts we've learned about, like `state`. Then, we're going to add some new functionality so we can use our application to take attendance for our students.
+
+## Refactor!
 
 **This diagram will drive our overall approach for accomplishing this goal:**
 ![nested components](images/nested-components.png)
 <!-- https://drive.google.com/open?id=1xq5jaCrI7FGp6PG1gr-bYE1ZTvPb5PxZ -->
 
-
-## Load data from `props`
-
-First, we want to use the Student component we already created, but we want to allow this to load data dynamically.
-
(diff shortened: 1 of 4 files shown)


React/events.md
```diff
@@ -2,85 +2,88 @@
...
 - Use an event handler to handle a change on a component's element
-- Modify state via an event handler
+- Modify `state` via an event handler
 
 ### Overview
 React supplies support on many different events. We'll focus on a few of the common events, keyboard and click, but just know that you should reference the [full list](https://reactjs.org/docs/events.html#supported-events) if you want more information.
 
 ### Context
-In this lecture, we are going to create a thrilling application that will make us happy by adding SPLATS to our page.
+In this lecture, we are going to expand upon the `DisplayName` component that we created in the [state lecture](state.md). This was a great component that we created that was storing a name and a boolean in the state. We used the boolean in the render function to determine what to display.
```
Hamled: In the state lecture we called the component `NameDisplay`.

React/creating-components-2.md
```diff
@@ -2,81 +2,73 @@
...
+1. Spend a few minutes now going back to the [creating components](creating-components.md) notes to see how to create the new component file with the basic class, imports and exports you need.
 
-Move the code that renders `Student` from `App` to `StudentCollection`. Note: You will also need to move and update the `import` statement.
+1. Move the code that renders each `Student` from `App` to `StudentCollection`. Note: You will also need to move and update the `import` statement.
 
-Import and render `StudentCollection` now from the `App` component instead.
+1. Import and render `StudentCollection` now from the `App` component instead. Note: You will need to add an import statement.
 
 ## Use `state` for data
+One way that we often set initial `state` data in a component, is by taking data passed in through `props` (from the parent) and setting that to `state` in our constructor. Let's see an example!
```
Hamled: It doesn't look like we show an example of a component's constructor function taking (or using) `props` as its parameter.

It's probably best to just mention it here even if we can't show an example (it doesn't seem like there's an easy way to provide an example of that here).

kariabancroft: I think we only need that in a functional stateless component, so maybe we should just include it there

droberts-ada:
LGTM, I think. I've got a PR against this branch with some fixes to the code in the notes - please merge that here before merging this to master.

PR #425, opened 2018-05-28
Cm/prop types

CheezItMan:
## Description
A small resource on `propTypes`.  

## Questions for the Author

## Feedback Requested


React/prop-types.md
```diff
@@ -0,0 +1,82 @@
+# Prop Types
+
+## Learning Goals
+By the end of this lesson you should be able to:
+
+- Explain the use of `PropTypes`
+- Use `PropTypes` to define and document the props a component can take
+
+You may have noticed this message from your JavaScript linter.  "____ is missing from props validation (react/prop-types)."  The linter is encouraging you to use a feature of React called Prop-types in which you list the props a component takes and their types.
```
kariabancroft: Can we make the "Prop-types" either all lowercase or camel case?

React/prop-types.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+The above block defines two props.  They both must be Strings and are required to use a `Student` component.  If you try to use the `Student` component without a `name` prop you will get this error message:
+
+![Error Msg](images/missing-prop.png)
+
+If a component has optional props you can simply leave off the `isRequired` attribute.
+
+**Exercise** Add another prop to the `Student` component called `age`.  This component should be _optional_.  Update the `Student` class' `propTypes`.
```
kariabancroft: You could also add the "present" attribute which we're going to use in the creating components lecture as optional boolean?

React/prop-types.md
```diff
@@ -0,0 +1,82 @@
...
+
+![prop types error](images/prop-types.png)
+
+In this lesson we will define `propTypes` in our React components which will provide some type-checking and documentation of our React component props.
+
+## Student Component Example
+
+Our Student component takes a name & an e-mail as props.  We can define the properties our `Student` component takes by adding a `static` `propTypes` attribute.
+
+First you need to `import PropTypes from 'prop-types';` to import the package.  Then add a static `propTypes` attribute to your class like below.
```
kariabancroft: I feel like there is not quite enough context about where exactly this should go in the component class. Could you add that detail?

React/prop-types.md
```diff
@@ -0,0 +1,82 @@
...
+
+```javascript
+
+static propTypes = {
```
kariabancroft: Do we also want to include the syntax they may see where you put this below the component class definition like:
`MyComponent.propTypes`?

CheezItMan: Updated to include it

React/prop-types.md
CheezItMan: Added

React/prop-types.md
CheezItMan: Updated

PR #436, opened 2018-06-08
JS Ecosystem lecture notes

Hamled:
## Description
This is the lecture notes for a lecture on the modern JS build tools ecosystem.

## Feedback Requested
Is this too detailed?

I started explaining the various tools and quickly realized that the students would need some kind of understanding of a build system and the idea of building a software project in the first place, since we basically don't need to do it with Ruby or Rails projects.

But I may have ended up going into too much detail when explaining that background stuff, so let me know if it should be cut down (and what parts you think should be cut if you have any suggestions).

### DIFF
--- a/10-JavaScript/js-ecosystem.md
+++ b/10-JavaScript/js-ecosystem.md
@@ -1,0 +1,46 @@
+# JavaScript Ecosystem
+The term "JavaScript ecosystem" is often used to refer to the array of different tools and configuration systems used for implementing a _build system_, generally for client-side projects including not only JavaScript code but also HTML, CSS, images and other data.
+
+## Learning Goals
+- Understand the purpose of a build system for client-side JS projects
+- Understand the role of the following specific tools within a build system:
+    - [npm](https://npmjs.com/)
+    - [webpack](https://webpack.js.org/)
+    - [Babel](https://babeljs.io/)
+    - [webpack-dev-server](https://github.com/webpack/webpack-dev-server)
+- Explore a complex build system setup by "ejecting" the build configuration from a `create-react-app` project
+
+## What is a build system?
+A build system for a software project is set of one or more programs (generally referred to as tools) that are usually connected by "script" code (written in a general programming language or one specific to a particular build system) which directs those tools to transform the project's source code and data files into a complete, deployable "package" for end users.
+
+An example of one of these end-user packages is the [ZIP file](https://atom.io/download/mac) that you would download when installing Atom. Inside of that compressed archive was an `Atom.app` directory that represents the _built_ version of the Atom project. If you extract the archive and then right click on the Atom file it creates and select "Show Package Contents" you can explore what files are included and see that they are very different from what's in the [project's GitHub repo](https://github.com/atom/atom) (it is not important to understand the files, just be aware of the difference).
+
+The need for a build system and the specifics of what it does are highly specific to each project and are often impacted by technology choices such as programming language, libraries used, and in particular the environment in which the program will be run by end users.
+
+## Why do we need a build system for our client-side JS projects?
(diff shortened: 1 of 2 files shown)


10-JavaScript/js-ecosystem.md
```diff
@@ -0,0 +1,140 @@
...
+- Understand the purpose of a build system for client-side JS projects
+- Understand the role of the following specific tools within a build system:
+    - [npm](https://npmjs.com/)
+    - [webpack](https://webpack.js.org/)
+    - [Babel](https://babeljs.io/)
+    - [webpack-dev-server](https://github.com/webpack/webpack-dev-server)
+- Explore a complex build system setup by "ejecting" the build configuration from a `create-react-app` project
+
+## What is a build system?
+A build system for a software project is set of one or more programs (generally referred to as tools) that are usually connected by "script" code (written in a general programming language or one specific to a particular build system) which directs those tools to transform the project's source code and data files into a complete, deployable "package" for end users.
```
kariabancroft: "is a set of"

Hamled:
@kariabancroft Yeah, I'm not sure how best to approach that.

If their goal is to simply include a third-party library for their project, that can generally be done without ejecting (e.g. I added the moment library to the react-timeline and react-chatlog apps that way).

On the other hand, if they need to customize their build process by adding some specific webpack loader or something else, that _would_ require ejecting. In that case the specific guidance gets a lot more complicated because the webpack config is already so complicated -- I'm not even sure how I would go about adding stuff into it.

Hamled:
@tildeee I added a new definitions table and linked to it from the learning goals.

Let me know if you think I should re-arrange the content or otherwise do something to call out that the definitions and overview on build systems is the core of what they should be learning.

10-JavaScript/js-ecosystem.md
```diff
@@ -0,0 +1,153 @@
...
+
+  However, this approach has some downsides. Specifically, the browser is loading a _file_ URL which looks like `file:///Users/username/some/directory/to/a/file.html`, which is treated differently from an _http(s)_ URL. For large or potentially complex projects it's important to mimic the actual end-user experience as closely as possible so we really want the browser to access our site through HTTP just as it would any other website on the internet.
+
+  In order to accomplish this we need to run a web server which will actually provide our HTML, CSS, JavaScript, and other files from a given URL (generally `localhost:3000` or similar). That is the purpose of the webpack-dev-server tool, specifically built to integrate with a webpack-based build system.
+
+  In very complex applications the developers might modify this "dev server" to have custom behavior akin to an API, but for our purposes we just need to run it with the default setup. In this case the dev server can most closely be compared to running `rails server` for our Rails projects.
+
+## Create React App
```
tildeee: Starting here, this section is for the following learning goal:
- Explore a complex build system setup by "ejecting" the build configuration from a `create-react-app` project

How would you feel if before this header, we had the following top-level header above it:
`# Explore A Complex Build System Setup through Create React App`

?

That might help structure a student's expectation for what to get out of this section

10-JavaScript/js-ecosystem.md
```diff
@@ -0,0 +1,153 @@
...
+
+Lastly, while the above webpack config from Create React App may have seemed really intense and unapproachable, rest assured that it is the result of many, many developers' efforts to build and maintain that config -- and to make it useful for as broad an audience as possible.
+
+In your personal projects, or your projects for work, it's very likely you can get away with a simpler configuration. And generally you will build that config piece-by-piece over time, adding in one new tool at a time when you end up needing it. Creating your build system in this incremental approach is vastly more approachable than starting with a huge one such as CRA's and trying to learn it all at once.
+
+## Definitions
+There's a lot of detail in the sections above, and while that can be helpful it's always good to have a quick reference. Consider the table below to be the core set of terms you should understand from this resource.
+
+| Term | Definition |
```
CheezItMan: 👍

PR #441, opened 2018-07-23
Update to Minitest

CheezItMan:
## Description
This is an update to testing with minitest.  It's needed for a couple of reasons. 
-   We are now introducing testing prior to classes and thus our example doesn't work anymore.
-   This should also be a smaller lesson, we're leaving out `before` blocks and only testing a single method.

The idea would be for them to do the textbook lesson as a code-along with some small partner practices and then a brief (30 minute) lesson where they practice TDD.  

We can then after introducing classes teach testing objects with the `Card` and `Deck` lesson.  

- [ ] Take a look at the Card & Deck lesson

## Feedback Requested


CheezItMan:
Since Rake is mostly a copy-paste thing which leads directly into the project and running the tests, I'd left it in.  It usually only takes 5-10 minutes.  

We usually teach `let` around week 4, I think, or around the Hotel week.

PR #447, opened 2018-08-12
Cm/intro to exceptions

CheezItMan:
## Description
A short(ish) intro to exceptions focusing on just what they are and how to raise your own exceptions (with `ArgumentError` as the example).  

## Todos
- [ ] Once merged in I'll add to the calendar
- [ ] Still need to update weekly learning goals.  

## Feedback Requested
Please take a look


01-ruby-fundamentals/raising-exceptions.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+This stack trace is more specific about the problem and prevents the application from going further with invalid arguments.
+
+## Why Raise ArgumentError Exceptions?
+
+By checking the parameters of a method for valid values & types we prevent unexpected output.  We could add to our method by adding a check to make sure that `number_of_people` is 1 or more people.
+
+```ruby
+unless order_items.class == Array && number_of_people.is_a? Numeric && number_of_people > 0
```
droberts-ada: Type checking again

droberts-ada:
Great work overall - I think this will be a really valuable resource.

PR #451, opened 2018-08-31
Testing With Guard resource

Hamled:
## Description
This adds a resource that explains what Guard is, gives simple installation instructions, and also has a deeper explanation of the Why for using a test watcher like Guard.

There is also a quick overview of the `Guardfile` configuration that we use for Ada projects.

## Todos
- [ ] We should figure out where to put this in the schedule. Only the installation instructions are needed for students to be able to use Guard for the Hotel project, and that should take probably 15 minutes or less to get set up.

## Feedback Requested
The final two sections (explaining why to use Guard) are largely off-the-cuff or stream of consciousness style, so if anyone has suggested wordsmithing or clean-up I can do for that, please let me know!


Hamled:
[Week 5: Test watcher for minitest](https://trello.com/c/V77O0cFM/982-week-5-test-watcher-for-minitest)

PR #452, opened 2018-09-04
Putting in Dan's fix for minitest spec helper

CheezItMan:
Just an example of Dan's code to get Guard to work with minitest-reporters


Hamled:
@CheezItMan I'm not sure why this file was updated to include an example `spec_helper.rb` file, as that is not related to the content of this resource.

The setup for Guard (including the `Guardfile` itself if students want to set that up) does not require any changes to `spec_helper.rb`, and that file should probably be introduced earlier in a separate resource.

I see that we have an example `spec_helper.rb` in [the code coverage resource](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/00-programming-fundamentals/code-coverage-and-simplecov.md) but that's also probably no the ideal location.

Perhaps it should go in [intro to automated tests](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/00-programming-fundamentals/intro-to-automated-tests.md)? Not sure anymore what is ideal since we've switched around projects, so I'm not sure which one is the first that includes a scaffold for automated testing.

PR #455, opened 2018-09-10
HTML forms: Minor changes to input and label elements

Hamled:
As it turns out, `input` elements are empty, so we can't put descriptive text within them (even if the browser will make it work). Instead we need to use labels for that in all cases.

Additionally, labels connect to input elements by matching the `for` and `id` tags, not `for`/`name`.


droberts-ada:
Today I learned...

PR #456, opened 2018-09-10
Add Guard to Rails templates

Hamled:
## Description
This adds necessary setup files for Guard to our Rails templates (both the initial and advanced versions).

## Feedback Requested
I've tested this out with a new Rails project, and it seems to work fine. Feel free to test it out with any of our existing project scaffolds!


Hamled:
@CheezItMan Thanks! I'm gonna merge it in based on that confirmation.

PR #459, opened 2018-09-13
Suggested changes to Rails forms lecture

Hamled:
## Description
This started off with some review comments and turned into a much more substantial modification.

The overall impact of the changes is to introduce Rails forms more slowly, and to simplify earlier examples a bit.

I also made a choice to briefly mention the hidden input tags generated by Rails, and then leave them out of further code samples in order to minimize distractions.

## Additional changes required
I would highly suggest splitting this into two resources now, by splitting out all of the ActiveRecord related content.


Hamled:
[Week 7: Convert from form_for to form_with](https://trello.com/c/5OTxQlCA/799-week-7-convert-from-formfor-to-formwith)

PR #460, opened 2018-09-14
Practice Exercise on Testing

CheezItMan:
## Description
I'm trying to build a few practice exercises involving:

1.  Writing code using loops and arrays, specifically processing a collection of objects
1.  Reading Tests and writing code to match
1.  Writing Tests 

I'll be adding to this today & tomorrow, but this is the first thing to look at. It covers writing tests and reading the tests of another student, and a bit of work with arrays. 

I'd like to see if I we could pair some of our struggling students to do these exercises because I think they might benefit from working together on something small. 

## Todos
- [ ] Exercise specifically using arrays of objects and loops to process them
- [ ] Exercise on reading tests and writing code to match (also using loops and arrays).

## Feedback Requested
Ideas are **VERY** welcome

### DIFF
--- a/02-intermediate-ruby/exercises/more-testing-practice.md
+++ b/02-intermediate-ruby/exercises/more-testing-practice.md
@@ -13,14 +13,14 @@ The goal of this set of exercises is to:
 
 ## Flatten Array Problem Description
 
-This method should take a nested list and return a single flattened list with all values except nil/null.
+This method should take a nested list and return a single flattened list with all values except nil.
 
-The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.
+The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil values.
 
 For Example:
 
 Ex 1:
-input: `[1,[2,3,null,4],[null],5]`
+input: `[1,[2,3,nil,4],[nil],5]`
 
 output: `[1,2,3,4,5]`
 
@@ -31,7 +31,7 @@ output: `['Peter', 'Paul', 'Mary']`
 
 ## Stage 1:  Write the Tests
 
-Write a set of tests to evaluate a method named `flatten_array` which takes a single array as a parameter and returns that array flattened like above.  It should also remove any null values.
+Write a set of tests to evaluate a method named `flatten_array` which takes a single array as a parameter and returns that array flattened like above.  It should also remove any nil values.
 
 
 ## Stage 3:  Test Review
--- a/02-intermediate-ruby/exercises/more-testing-practice.md
+++ b/02-intermediate-ruby/exercises/more-testing-practice.md
@@ -43,6 +43,14 @@ With a classmate review your written tests and look for:
 *   Opportunities to DRY-up code
 
 
-## Stage 4:  Refactor and Solution
+## Stage 4:  Refactor
 
-Finally refactor your tests from your partner's suggestions and write a `flatten_array` method to satisfy your tests.  You may not use the Array class' built-in method to flatten the array.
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/exercises/more-testing-practice.md
```diff
@@ -0,0 +1,48 @@
...
+
+In this exercise you will look at a problem description and write a set of tests to evaluate a solution.  Then you will review the tests of another classmate and make suggestions for improvement.  Lastly you will write solution code to satisfy your tests.
+
+## Exercise Goal
+
+The goal of this set of exercises is to:
+
+*   Practice writing tests
+*   Review the tests written by another and look for:
+  -Testing all edge-cases
```
kariabancroft: These two bullets are not formatted correctly

02-intermediate-ruby/exercises/more-testing-practice.md
```diff
@@ -0,0 +1,48 @@
...
+## Flatten Array Problem Description
+
+This method should take a nested list and return a single flattened list with all values except nil/null.
+
+The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.
+
+For Example:
+
+Ex 1:
+input: `[1,[2,3,null,4],[null],5]`
```
kariabancroft: I think we should use `nil` if we're planning on giving this to them within the context of their time in Ruby. Then we could avoid any confusion with the nil/null section above.

02-intermediate-ruby/exercises/more-testing-practice.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Flatten Array Problem Description
+
+This method should take a nested list and return a single flattened list with all values except nil/null.
+
+The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.
+
+For Example:
+
+Ex 1:
```
kariabancroft: The space between the example header and input is flattened, so if you add another carriage return you should be able to see it

02-intermediate-ruby/exercises/more-testing-practice.md
```diff
@@ -0,0 +1,48 @@
...
+output: `[1,2,3,4,5]`
+
+Ex 2:
+input: `['Peter', ['Paul', [ 'Mary' ]]]]`
+
+output: `['Peter', 'Paul', 'Mary']`
+
+## Stage 1:  Write the Tests
+
+Write a set of tests to evaluate a method named `flatten_array` which takes a single array as a parameter and returns that array flattened like above.  It should also remove any null values.
```
kariabancroft: Do you think it would be worthwhile to add a "Think about" section where we provide a few basic prompts to help them think about the types of things they should be testing (without giving it away)?

02-intermediate-ruby/exercises/more-testing-practice.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Stage 3:  Test Review
```
kariabancroft: Should this be Stage 2 instead or as we missing a set of steps you want to include?

02-intermediate-ruby/exercises/more-testing-practice.md
```diff
@@ -0,0 +1,48 @@
...
+With a classmate review your written tests and look for:
+
+*   Edge-cases uncovered
+*   Return values tested correctly
+*   Opportunities to DRY-up code
+
+
+## Stage 4:  Refactor and Solution
+
+Finally refactor your tests from your partner's suggestions and write a `flatten_array` method to satisfy your tests.  You may not use the Array class' built-in method to flatten the array.
```
kariabancroft: I'd love to see as part of this step a reflection on whether or not the tests that they came up with ended up covering the code that they decided to write. I think that is also an interesting this to examine.

CheezItMan:
@kariabancroft Thanks, I've revised it into more of an exercise in identifying where code will fail the tests, rather than solving the problem.  I'll add a solution later for those who want to see one.

PR #463, opened 2018-09-17
Small AR Migrations Practice

CheezItMan:
## Description
Just a few additions to the migrations lecture to get them to try the commands.  The models lecture actually has quite a bit of terminal practice for'em.  

## Feedback Requested
Let me know if this is useful or not.

### DIFF
--- a/08-rails/active-record-migrations.md
+++ b/08-rails/active-record-migrations.md
@@ -10,10 +10,10 @@ Active Record (AR) is an important piece of the Rails approach to web applicatio
 __Question: What is the role of _models_ in the MVC pattern?__
 
 ## A Few Terms
--  A Database _schema_ is the structure of a database.  It's the logical view of the tables and how they are related to one another.  
+-  A Database _schema_ is the structure of a database.  It's the logical view of the tables and how they are related to one another.
 -  A _migration_ is a change to that database schema.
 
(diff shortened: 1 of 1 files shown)


08-rails/active-record-migrations.md
```diff
@@ -91,6 +91,21 @@ end
 
 Once we have the migration created, we must again run `rails db:migrate` for the migration to take effect.
 
+## Practice
+
+In our sample Rails project perform the following actions:
+
+1.  Create a new model named: `Author`, an author should only have a `name` field (String).  Remember to run your migration with `rails db:migrate`
+2.  Then add a migration to add a `publisher` field to the model.
```
kariabancroft: I usually try to say "Create a migration which will add a...." rather than "add a migration" to indicate the way that it needs to create the files rather than adding to an existing file.

08-rails/active-record-migrations.md
```diff
@@ -91,6 +91,21 @@ end
...
+
+Then run `rails db:migrate:status` to verify that the migrations have run successfully.
+
+You can then destroy your `Author` model with:
+
+```bash
+rails destroy model Author name:string publisher:string
```
kariabancroft: I will pretty much never have them do this command, so i'm wondering if we should just avoid adding the author altogether. They should run exactly the model and migration commands in task list (i think?)

PR #470, opened 2018-10-02
add validations resource

kariabancroft:
## Description
Create a resource to describe the difference between client and server side validations. Based on the Trello card, I think the intention of something like this would be a quick HW read rather than an in-class exercise or lecture. As such, please see my two questions below on if it should be added anywhere else. 

## Questions for the Author
- [ ] Have you updated the weekly learning goals in the pedagogy resource - **should this be added as a learning goal to a specific week?**
- [N/A] Have you added links to the source (lucidchart, draw.io) for any diagram

## Todos
- [ ] QUESTION: Should I add this to the overall README for this module?

## Feedback Requested
Does this work for a quick HW reading?

### DIFF
--- a/09-intermediate-rails/validations.md
+++ b/09-intermediate-rails/validations.md
@@ -1,0 +1,36 @@
+# Client Side vs Server Side Validation
+
+## Learning Goals
+
+By the end of this lesson, students should be able to...
+- Define validation
+- Define client-side validation
+- Define server-side validation
+- Understand the differences between the two
+- Understand the benefits of both types of validation
+
+### Validation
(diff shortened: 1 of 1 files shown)


kariabancroft:
[Week 11: Client side validation vs server side validation](https://trello.com/c/g19ciiS7/768-week-11-client-side-validation-vs-server-side-validation)

kariabancroft:
[Week 10: Add testing requirements to MediaRanker](https://trello.com/c/jg59iG0w/964-week-10-add-testing-requirements-to-mediaranker)

PR #475, opened 2018-11-03
Add Resource on API Testing

CheezItMan:
## Description
Previously we'd just had the tests pre-set in the ada-pets repository.  Do we need a separate lesson or can they get the info from the code?  Regardless I think having a small textbook resource would be helpful.

## Questions for the Author
- [ x ] Have you updated the READMEs with links to new or moved resources?

## Todos
About to submit an update to ada-pets.

## Feedback Requested


09-intermediate-rails/testing-api-controllers.md
```diff
@@ -0,0 +1,91 @@
...
+  end
+end
+```
+
+In the tests above we verify that the JSON response contains an array with the proper number of elements, and that each element of the array is a JSON hash with the required fields.
+
+## Exercise:  Write tests for the show action
+
+Now with your partner, write tests for the `show` action.  If you get stuck you can find a solution [here](./code_samples/pet_controller_test.rb).
+
```
droberts-ada: Might be good to add a question (or even another exercise section) about what testing would look like for `create`.

PR #483, opened 2018-11-22
React Week 1 Changes

droberts-sea:
## Description
Start with functional components. Includes a large number of small clarifications and updates.

## Questions for the Author
- [x] Have you updated the weekly learning goals in the pedagogy resource?

## Todos
- [ ] End-of-week project ([Radio Lovelace](https://github.com/AdaGold/radio-lovelace))
- [x] Intro to classes #482

## Feedback Requested
Probably the best approach would be to follow the order in the README.

I've tested all the code myself, so I'm mostly interested in feedback on prose and whether or not we do a good job of hitting the learning goals.

Also, I will be OOO starting Monday 11/26, so if you see something that needs fixing, fix it!

### DIFF
--- a/React/prop-types.md
+++ b/React/prop-types.md
@@ -23,7 +23,7 @@ const Student = (props) => {
   // Component functions always return JSX
   return (
     <div>
-      <h3 className={props.present ? 'present': 'absent'}>
+      <h3>
         {props.fullName}
       </h3>
       <p>{props.email}</p>
@@ -33,12 +33,13 @@ const Student = (props) => {
 
 Student.propTypes = {
   fullName: PropTypes.string.isRequired,
-  email: PropTypes.string.isRequired,
-  present: PropTypes.bool,
+  email: PropTypes.string,
 };
 ```
 
-The above block defines two props.  They both must be Strings and are required to use a `Student` component.  The attribute `present` however is an optional boolean prop.  If you try to use the `Student` component without a `fullName` prop you will get this error message:
+Wait a second - what's with that syntax? We just attached an attribute (`propTypes`) to a function (`Student`). Since functions are first-class objects in JavaScript, this is a totally valid thing to do. It even has a name: `propTypes` is a "static attribute" of `Student`.
+
+The above block defines two props. They both must be Strings. `email` is optional, but `fullName` is required to use a `Student` component. If you try to use the `Student` component without a `fullName` prop you will get this error message:
 
 ![Error Msg](images/missing-prop.png)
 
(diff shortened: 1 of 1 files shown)


React/nested-event-handling.md
```diff
@@ -1,66 +1,89 @@
...
+## Refactor to a Stateful Component
 
-1. Spend a few minutes now going back to the [creating components](creating-components.md) notes to see how to create the new component file with the basic class, imports and exports you need.
+### Lifting State Up
 
-1. Move the code that renders the list of `Student`s from `App` to `StudentCollection`. Note: You will also need to move and update the `import` statement.
+Currently there is no way for us to change the student data. If we want to be able to change whether a student is present, we'll need to keep track of that as state in our application.
 
-1. Import and render `StudentCollection` now from the `App` component instead. Note: You will need to add an import statement.
+**Question:** Where should our application's state live? What are our options?
```
tildeee: I think that our students probably won't be able to even hypothesize an answer for this as they know they don't know the implications of where state should live -- As a lecturer I'd probably make this a rhetorical question at the moment.

React/nested-event-handling.md
```diff
@@ -1,66 +1,89 @@
...
-        }
-      ];
+      {
+        fullName: "Ada Lovelace",
+        email: "ada@lovelace.uk",
+        isPresent: false
+      },
+      {
+        fullName: "Katherine Johnson",
+        email: "kat@nasa.gov",
```
tildeee: v good sample data thank you

React/nested-event-handling.md
```diff
@@ -1,66 +1,89 @@
...
-        },
-        {
-          fullName: "Grade Ada",
-          email: "grace@ada.co",
-          isPresent: false
-        }
-      ];
+      {
+        fullName: "Ada Lovelace",
+        email: "ada@lovelace.uk",
```
tildeee: yes good

React/prop-types.md
```diff
@@ -90,7 +86,7 @@ A `defaultProp` attribute like above ensures that every time the given props hav
 You can also create a custom validator which you can look up along with more examples of using `PropTypes` on the [PropTypes github page](https://github.com/facebook/prop-types).
 
 ## Key Takeaway
-`propTypes` is a static attribute you can define in a React component class which lists the props the component can take along with their types.  Mismatches or missing props will generate error messages in the browser console.  PropTypes help you document the props a component can take and help prevent the wrong types of props being passed accidentally.
+`propTypes` is an attribute you can define in a React component which lists the props the component can take along with their types.  Mismatches or missing props will generate error messages in the browser console.  PropTypes help you document the props a component can take and help prevent the wrong types of props being passed accidentally.
```
tildeee: So... students won't know what `static` is in terms of JavaScript officially by this point? We have definitions for it in Ruby but I think we still call them class methods and never static, and we have a definition of `static` methods for JavaScript classes on Wednesday (this lesson is slated for Tuesday)... Also, a static method on a class isn't quite the same as what's happening here...

Not sure what to do except maybe have a one sentence explanation somewhere.

droberts-ada: Aside from just being static, it's weird b/c we're attaching a property to a function, which they haven't seen before.

tildeee: "functions are objects and have properties"?

PR #484, opened 2018-11-24
Class Components

CheezItMan:
## Description
Added Class Components lecture.  I tried to keep this as much to the syntax as possible because the state lecture provides the answer to **why** use classes.  

## Questions for the Author

## Feedback Requested


droberts-ada:
What is this commit history? Did you branch off the wrong place?

React/class-components.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Questions:**
+
+1. What is the superclass for `Student`?
+1. What is accomplished with `super(props)`?
+
+### Is it `extends React.Component` or `extends Component`
+
+You may have noticed that `App` inherits from `Component` while the `Student` inherits from `React.Component`.  They are both inheriting from the same class, but in `App.js` we have _destructured_ the `React` object with `import React, { Component } from 'react';`.  The `{ Component }` part of the line is the equivalent to:  `const Component = React.Component`.  It's a way of assigning elements of an object to individual variables.  
+
```
droberts-ada: I like this explanation!

React/class-components.md
```diff
@@ -0,0 +1,110 @@
...
+We can see an example by re-writing our Student component as a class.
+
+```javascript
+// components/Student.js
+import React from 'react';
+
+class Student extends React.Component {
+  constructor(props) {
+    super(props);
+  }
```
droberts-ada: Unless you've got some specific behavior you intend to implement, I would probably not override the constructor here. It makes the code more complex, and doesn't really add to the discussion.

React/class-components.md
Hamled: I also think this is a good explanation of destructuring within an `import` statement (I think technically it's separate from the destructuring used in other parts of the JS syntax).

However, I'm not sure this is a great place to explain this concept. It's embedded like 2/3rds of the way through this lecture, and the only connection it has to the rest of the lecture is that we've chosen to use the non-destructuring syntax in our code snippets here.

If we instead agreed to always use the `import { React, Component} from 'react';` version instead, then we wouldn't have any discrepancy in how we write the parent class, and we could leave this explanation out entirely.

React/class-components.md
```diff
@@ -0,0 +1,110 @@
...
+## When to use a Functional or Class Component
+
+At Ada we start writing functional components because we believe they are the clearest to understand and easiest to get started writing JSX.  Furthermore, most of the time you will likely start a component as a functional component, even if you later change it into a class because you need the additional functionality.  
+
+Functional Components
+
+- Involve less complicated and lengthy syntax.
+- Can be understood more quickly because of their short, declaritive nature.
+- Are easier to test and debug because given a set of props, they **always** return the same JSX.
+- Provide better **performance** because they do not inherit functionality provided by the Component class.
```
Hamled: This doesn't seem true... where did this fact come from?

PR #501, opened 2019-02-08
Add resource on common inheritance idioms

droberts-ada:
## Description
We talk about the syntax of inheritance, but it would be good to introduce some of the vocabulary and design patterns that reflect how inheritance is really used.

This is work that I came up with while thinking deeply about inheritance for [this blog post](https://github.com/droberts-ada/blog-drafts/pull/1). I think it will make OO RideShare go a little smoother, and provide a slick transition into Rails and React some day.

This will be targeted for Tuesday or Wednesday of whatever week contains RideShare (previously 4, now 5).

## Questions for the Author

## Todos
None

## Feedback Requested
- Shruti, I added you because you reviewed the blog post!
- The example at the end is a lot of code. However, I'm having trouble coming up with good comprehension questions. Any ideas about what to do about all this?

### DIFF
--- a/02-intermediate-ruby/inheritance-idioms.md
+++ b/02-intermediate-ruby/inheritance-idioms.md
@@ -1,0 +1,21 @@
+# Inheritance Idioms
+
+## Learning Goals
+
+By the end of this lesson, students should be able to...
+- Define the term _abstract class_
+- Identify and implement _template methods_
+
+## Introduction
+
+There are many common patterns around how inheritance is used in real software projects that are not obvious from syntax alone. Today we will discuss two of these: _abstract classes_ and _template methods_.
+
+### What is an Idiom?
+
+### Abstract Classes
+
+### Template Methods
+
+## Summary
+
+## Additional Resources
--- a/02-intermediate-ruby/inheritance-idioms.md
+++ b/02-intermediate-ruby/inheritance-idioms.md
@@ -4,12 +4,17 @@
 
 By the end of this lesson, students should be able to...
 
-- Define the term _abstract class_
+- Define the following terms
+  - Abstract class
+  - Template method
+  - Polymorphism
+- Write code that extends an _abstract class_
 - Identify and implement _template methods_
+- Take advantage of _polymorphism_
 
 ## Introduction
 
-There are many common patterns around how inheritance is used in real software projects that are worth studying explicitly. Today we will discuss two of these: _abstract classes_ and _template methods_. Both these patterns are used to better tie together parent and child classes.
+There are many common patterns around how inheritance is used in real software projects that are worth studying explicitly. Today we will discuss three of these: _abstract classes_ and _template methods_, and _polymorphism_.
 
 ### What is an Idiom?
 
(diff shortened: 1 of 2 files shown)


02-intermediate-ruby/inheritance-idioms.md
```diff
@@ -0,0 +1,255 @@
...
+<details>
+<summary>Click for the answer</summary>
+Object, CsvRecord from the project
+</details>
+
+Returning to our example from [the previous lesson](object-inheritance.md), `Property` would be a good candidate for an abstract class. Instead of creating a generic `Property`, it would make more sense to always create an `Apartment`, `Condo`, or `House`.
+
+Some languages, particularly statically typed languages like C++ and Java, have built-in support for abstract classes. Ruby does not - any class that is not instantiated is abstract, and the way to make class abstract is to not instantiate it.
+
+### Template Methods
```
kaidamasaki: In this section do we want to mention that these are a type of abstract method?

02-intermediate-ruby/inheritance-idioms.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+## Summary
+
+Vocabulary
+
+ Term            | Definition                                                 
+-----------------|------------------------------------------------------------
+ software idiom  | A common pattern used by programmers. Not always obvious from the syntax of the language. 
+ abstract class  | A class that is not instantiated, but instead serves as a parent class.
```
kaidamasaki: Maybe "but instead serves only as a parent class"?

02-intermediate-ruby/inheritance-idioms.md
```diff
@@ -0,0 +1,255 @@
...
+
+**An _abstract class_ is a class that is never instantiated**. Instead, it serves as the parent class for some number of child classes, which are instantiated.
+
+**Question:** What abstract classes have we seen so far?
+
+<details>
+<summary>Click for the answer</summary>
+Object, CsvRecord from the project
+</details>
+
```
shrutivanw: Somewhere in this section, it could be useful to call out: Why are abstract classes used? Maybe an example could help.
The abstract class provides a contract for all the derived sub-classes to follow. Without the contract being satisfied, the derived sub-classes cannot be instantiated.

_Aside_: I don't talk about abstract classes and templates. While talking about inheritance and polymorphism, I use the example of Car class that a Tesla class gets extended from. See slide 22 on https://drive.google.com/drive/folders/0B__DV26QHsH4dXJTRXlJc1oydGM

02-intermediate-ruby/inheritance-idioms.md
shrutivanw: Since the students would be making the call while designing the system, instead of "*is* not instantiated", could this be: "A class that can not be (or should not be) instantiated"?

02-intermediate-ruby/inheritance-idioms.md
```diff
@@ -0,0 +1,255 @@
...
+## Summary
+
+Vocabulary
+
+ Term            | Definition                                                 
+-----------------|------------------------------------------------------------
+ software idiom  | A common pattern used by programmers. Not always obvious from the syntax of the language. 
+ abstract class  | A class that is not instantiated, but instead serves as a parent class. 
+ concrete class  | A class that is instantiated. May or may not inherit from some other class.
```
shrutivanw: Similar to above, "A class that can be (or should be)" instantiated."?

02-intermediate-ruby/inheritance-idioms.md
```diff
@@ -0,0 +1,255 @@
...
+Vocabulary
+
+ Term            | Definition                                                 
+-----------------|------------------------------------------------------------
+ software idiom  | A common pattern used by programmers. Not always obvious from the syntax of the language. 
+ abstract class  | A class that is not instantiated, but instead serves as a parent class. 
+ concrete class  | A class that is instantiated. May or may not inherit from some other class. 
+ template method | A method that is intended to be overridden in a subclass.  
+ interface       | Some set of methods that go together. A class that has all those methods _implements_ that interface. 
+ polymorphism    | Treating instances of different classes in the same way. All those classes must implement the same interface.
```
shrutivanw: I like this page. I'm getting a little greedy here for adding more details. 

Could we add more information on:
- *runtime polymorphism*, which is what is described in the example here.
- *compile time polymorphism*, available in C++, C# and Java
- *interfaces*: There is a recurring theme of confusion around abstract class versus interfaces in Java, especially in the first two weeks of internship.
- *multiple inheritance*: Some languages support it, some don't. Java doesn't and encourages thinking more about the public interfaces the class should implement.

02-intermediate-ruby/inheritance-idioms.md
```diff
@@ -0,0 +1,255 @@
...
+
+puts "We have many marvelous properties avaiable!"
+
+puts "All figures assume a the minimum downpayment on a 30-year loan with a 4% interest rate"
+
+properties.each do |property|
+  # We don't care if the property is an apartment or a condo
+  puts
+  puts property.address
+  payment = property.monthly_payment(30, 0.04)
```
shrutivanw: Suggestion: An explicit call out here could help to share the power of polymorphism: at runtime, behind the scenes, the language will determine what type of concrete class, and which subclass the object `property` is and accordingly invoke the correct implemented method in that sub-class.

02-intermediate-ruby/inheritance-idioms.md
droberts-ada: I like the idea, but I think it's too much for this point in the curriculum (week 4).

PR #503, opened 2019-02-11
Add intro to APIs

droberts-ada:
## Description
Smash together intro to the web from HTML/CSS week, intro to the web from Rails week, Requests/Responses from Rails week, and intro to APIs from the old API week.

Intended as the auditorium lesson at the very beginning of week 6.

The end result is kind of long (definitely an hour worth of content), but there's some good interactivity in the middle. I think it will set us off on the right foot.

There's a fair bit of overlap with the requests/responses lesson from the intro to Rails week, but I think keeping both is valuable since the Rails lesson addresses it from a server perspective. The review / connection to previous learning will be good for them.

## Questions for the Author

## Todos
None

## Feedback Requested
Nothing special


02-intermediate-ruby/api-intro.md
```diff
@@ -1,7 +1,215 @@
...
+Verb             | Meaning
+---              | ---
+`GET`            | Send me some data
+`POST`           | Here is some data, please save it
+`PATCH` or `PUT` | Make a change to existing data
+`DELETE`         | Remove data
+
+There's quite a few default _verbs_ in the HTTP spec, though we (Ruby/Rails devs) mostly just use `GET`, `POST`, `PUT` (and `PATCH`), and `DELETE`.
+
+The _path_ tells the server what resource the client is interested in. In our example above this is the list of books at `/books`. Later we'll see it looking at individual books via the path `/books/3` (where 3 is the ID of the book in question).
```
tildeee: Since this is their first exposure to the idea of a "path," it might be interesting to give a small example before `/books` to relate the idea of a URL to a `path`

02-intermediate-ruby/api-intro.md
```diff
@@ -1,7 +1,215 @@
...
+- Whenever computers communicate with each other, they use a _protocol_
+  - All the work we do in this class will involve the HTTP protocol
+- There are two main types of resource on the web:
+  - _Websites_ are designed to be used by humans
+  - _APIs_ are designed to be used by other machines
+  - Browsers can consume both
+- A key pattern on the web is the _request/response cycle_
+  - Client makes a request, including HTTP verb, path, optional payload
+  - Server sends a response, including HTTP status code, optional payload
+
```
tildeee: You mentioned this before, but this lesson is pretty long/has a lot of important new information. It might be nice to have a series of high-level comprehension questions at the end here (before or after the Summary section. I like after the Summary section because it has a lot of the answers hehe).

For example, if I were giving this lesson/writing the questions, I would probably ask students to answer this with the people next to them:

- Describe the relationship between client and server-- if we write a Ruby CLI that makes a request to an API, which is the client and which is the server?
- What data type did the API `dog.ceo` (from the example above) give back (looking for JSON)?
- What are the HTTP Verbs?

If you want to pursue this, also... maybe just a series of Summary/Details tags? :D Then they would have the answers if we ran out of time in lecture.

PR #513, opened 2019-03-01
Update inheritance-idioms.md

kaidamasaki:
Add a few examples of English idioms.


kaidamasaki:
I thought these might help illustrate what an idiom is.

PR #516, opened 2019-03-05
Cm/let

CheezItMan:
## Description
An update to the `let` lecture.  The goals are to connect it to what they've learned before with `attr_reader` etc, and making the example code runnable with 2 files, `pet.rb` and `pet_spec.rb`.  I also took out unnecessary code.  

## Feedback Requested

### DIFF
--- a/02-intermediate-ruby/lazy-with-let.md
+++ b/02-intermediate-ruby/lazy-with-let.md
@@ -37,7 +37,7 @@ In most test cases we're creating a new identical instance of `Pet` to work with
 
 ## Solution 1:  Before Blocks
 
-One solution is to use Minitest's `before` block.
+One solution is to use Minitest's `before` block.  In Minitest a `before` block runs before each test case, which allows you to dry up common code in the arrange step of testing.
 
 ```ruby
 require 'minitest/autorun'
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/lazy-with-let.md
```diff
@@ -33,19 +56,27 @@ describe Pet do
 end
 ```
 
-So in the above example the `before` block runs before each test and provides an instance variable `@pet` which we can use in our tests.  The problem with before blocks is that they **always** run before each test case.  Even if we don't want to use `@pet` in a few test cases, like for instance if we want to try creating a Pet with no arguments the before block will still run.
+In the above example the `before` block runs before each test and provides an instance variable `@pet` which we can use in our tests.  This can often serve as a helpful way to DRY testing code.
+
+Unfortunately `before` blocks **always** run before each test case.  Even if we don't want to use `@pet` in a few test cases, like for instance if we want to try creating a Pet with no arguments the before block will still run.
```
droberts-ada: Could you talk about the connection b/w `before` and `describe`? Specifically, the idea that if you have multiple `describe` blocks or nested `describe`s, you can control where it happens a little tighter.

PR #540, opened 2019-04-03
Update testing in controllers

CheezItMan:
## Description
Before, when we did controller testing students knew about fixtures and validations.  Now we need to teach them controller testing without either of those.  So I've updated the lecture (and fixed the missing objectives) to match our new week 1 plans.  

## Feedback Requested

### DIFF
--- a/08-rails/testing-db-in-controllers.md
+++ b/08-rails/testing-db-in-controllers.md
@@ -77,15 +77,15 @@ The example above illustrates that the test can pass in a mock-params hash into 
 
 **Exercise** You also need to create a test in which the params are invalid or missing.  With a partner write another test in which the params are invalid.  We have an [example solution](code_samples/create_controller_test.rb)
 
-**Exercise** Similar to the `create` action tests, write tests to verify the correctness of the `update` action.  You should have at least 3 tests. When you finish you can [view our solution](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/08-rails/code_samples/update_controller_test.rb)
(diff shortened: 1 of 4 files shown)


08-rails/testing-db-in-controllers.md
```diff
@@ -49,21 +63,31 @@ describe "create" do
...
+    expect(book.author).must_equal book_hash[:book][:author]
+    expect(book.description).must_equal book_hash[:book][:description]
   end
 end
 ```
 
 The example above illustrates that the test can pass in a mock-params hash into the request with the `params: params_hash` argument.  Notice we verify that the number of books in the database increases and that the last book in the database has the correct title, author and description.
 
-**Exercise** You also need to create a test in which the params are invalid, violating validations.  With a partner write another test in which the params are invalid.  You can see an example [here.](code_samples/create_controller_test.rb)
+**Exercise** You also need to create a test in which the params are invalid or missing.  With a partner write another test in which the params are invalid.  We have an [example solution](code_samples/create_controller_test.rb)
```
droberts-ada: If the params are missing, will that cause a problem? Remember that we have no validations for this model.

CheezItMan: There wont' be a `:book` field in params and strong params should fail causing an exception.  We could still take this test out.

droberts-ada: Ah, I see. I like the idea then, it'll be good to have a test for a failure case.

Will we have introduced strong params at this point in the week? Looks like we don't cover that until Friday?

CheezItMan: I'll pull it out for Wednesday and add it back in on Friday.

PR #548, opened 2019-04-15
Ddr/db seeds smoothing

tildeee:
## Description
Smoothed out the db seeds lecture so we can use it in our "Logic in the Model" lecture


dHelmgren:
lgtm!

PR #572, opened 2019-05-17
split out fun with functions, prioritize small exercises

tildeee:
## Description
fun with functions is to be broken up into 4 small lessons with a lot of small exercises

## Questions for the Author
- [x] Have you updated the weekly assessment in Schoology?

## Todos
- [x] more splitting!

## Feedback Requested


10-JavaScript/fun-with-functions.md
```diff
@@ -1,34 +1,99 @@
-# Fun with Functions!
+# Intro to Functions!
 
 ## Learning Goals
-- Write and invoke some JavaScript functions
+
+- Intro to syntax (defining and calling), exercise, functions with parameters, exercise
```
kaidamasaki: "exercise" feels weird to be in the learning goals.

Also, I might make "defining", "calling" and "functions with parameters" as sub bullets.

10-JavaScript/fun-with-functions.md
```diff
@@ -1,34 +1,99 @@
...
+
+## Functions in JavaScript are like Methods in Ruby
+
+In JavaScript, we can use functions to define reusable behavior.
+
+This should _feel_ similar to methods in Ruby. They're not the same, but we can get into those details later. For now, we'll use functions to name and structure sections of code.
+
+Either way, the syntax for functions is a lot different in JavaScript than in Ruby.
+
+## Functions Are Objects, So We Assign Variables to Them
```
kaidamasaki:
```suggestion
## Functions Are Objects, So We Can Assign Them to Variables
```

10-JavaScript/fun-with-functions.md
```diff
@@ -1,34 +1,99 @@
...
+
+Unlike in Ruby, JavaScript requires you to type in parentheses if you want to invoke a function:
+
+```javascript
+bark();
+```
+
+### The Finer Details About Defining Functions
+
+When we defined the function, we gave it the variable keyword `const`. We could have used `let` or `var`, but in general, we don't need any variable named `bark` to be reassigned. Because we don't need `bark` to every be reassigned, it makes sense to make it a constant.
```
kaidamasaki: I'd like to avoid mentioning `var` more than we have to.

```suggestion
When we defined the function, we gave it the variable keyword `const`. We could have used `let`, but in general, we don't need any variable named `bark` to be reassigned. Because we don't need `bark` to ever be reassigned, it makes sense to make it a constant.
```

10-JavaScript/fun-with-functions.md
tildeee: lmao this was a note that i leftover for myself, oops, deleting now

10-JavaScript/fun-with-functions.md
tildeee: So I thought a lot about this... Idk, it's not _that_ important, but I really think of it as variable assignment that we assign variables to values, we don't assign values to variables........ idk idk but I would fight to keep as-is

kaidamasaki: I initially parsed this as talking about being able to add fields to functions.

10-JavaScript/first-class-functions.md
```diff
@@ -0,0 +1,241 @@
...
+  isComplete: false,
+  markComplete() {
+    console.log('Wow...');
+    console.log('This task is complete!');
+    console.log('Congratulations! You won!');
+  },
+}
+```
+
+We have given the object `task` the method `markComplete()` by:
```
kaidamasaki:
```suggestion
We have given the object `task` the **method** `markComplete()` by:
```

10-JavaScript/first-class-functions.md
```diff
@@ -0,0 +1,241 @@
...
+</details>
+<br/>
+
+Be aware that JavaScript's `this` keyword has some strange behavior. Many times it won't refer to quite what you'd expect. This can be one of the most frustrating things about JavaScript, especially for a beginner, and we'll have more to say about it later. For now, just know that it's a thing that might come up.
+
+## We Can Pass Functions as Arguments
+
+Remember how functions can take in arguments? We can define parameters, which is our way of defining inputs. Those arguments can be numbers, strings, other objects, and so on. We can read them, modify them, and return them as output.
+
+In JavaScript, functions are objects. That means **you can pass functions as arguments to other functions.**
```
kaidamasaki:
```suggestion
In JavaScript, functions are objects. That means **you can pass functions as arguments to _other functions_.**
```

10-JavaScript/first-class-functions.md
```diff
@@ -0,0 +1,241 @@
...
+  <summary>
+    Check your answers here!
+  </summary>
+
+  1. 1 parameter named `num` that is a number
+  1. 1 parameter named `operation` that is a function
+  1. `doMath(doubleNum)`. We pass in the `doubleNum` function as an argument.
+  1. Because `operation` is a function, we _invoke_ the `operation` function with `let result = operation(i);`
+  1.
+      ```bash
```
kaidamasaki:
```suggestion
      ```
```

10-JavaScript/functional-expressions.md
```diff
@@ -0,0 +1,29 @@
...
+A functional declaration looks like this. Note the lack of `const` and a semicolon, and the different word order.
+
+```javascript
+// functional declaration - this is the wrong way
+function foo(bar) {}
+```
+
+In many cases the differences are nominal, but every once in a while they'll bite you. The vast majority of the time, a functional expression is what you want. In other words, **always define your functions with the `const` keyword**. The declaration style (no `const`) should be avoided unless you have a good reason to use it.
+
+If you're interested in more details, [this StackOverflow answer](http://stackoverflow.com/questions/3887408/javascript-function-declaration-and-evaluation-order) is one of the best we've seen in describing the difference and when it matters. We don't expect you to be able to recite the nitty gritty details, but we do expect you to be able to identify the two types of function and tell us which is correct.
```
kaidamasaki: I think we should maybe extol the virtues of `const` here instead of being like "you should type the names twice because of some edge cases that probably won't come up" because otherwise students will be tempted to use the old style syntax.

tildeee: you're right, that's a way better attitude!

10-JavaScript/anonymous-and-for-each.md
```diff
@@ -0,0 +1,130 @@
...
+Each of these anonymous function callbacks should take in 1 number, and return either a number or a string.
+
+Ideas for anonymous functions to write include:
+- Cubing one number
+- Formatting the input like it was currency
+- Multiplying it by 1 million and then formatting it like currency and then adding text saying that it belongs to the Ada instructors
+
+Run your script to make sure it works!
+
+## Practical Example: `forEach` Loops
```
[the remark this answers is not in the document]
tildeee: We end up teaching `map` as part of our lesson on refactoring to arrow functions, so I'll leave it for that day!

PR #583, opened 2019-06-13
Make this compatible with a functional Student component

kaidamasaki:
Not sure if we want to do this but it makes sense for `<Student>` to be functional since it doesn't have state.


kaidamasaki:
We also could do this by defining the callback within the body of the functional component and close over props that way.  (Which I think might be nicer.)

kaidamasaki:
Yeah, I think the other way is cleaner.

PR #586, opened 2019-06-25
updated to include closures

dHelmgren:
## Description
I've updated the lecture on event handling to show how state can move from one component to another, _and_ shown students how to clean up the child component along the way.

Best practice for review will be to read the file as a whole, just to make sure all the connective tissue is still good.


dHelmgren:
[Week 18: Fix-Up Event Handling With Nested Components](https://trello.com/c/onH9DVAH/1569-week-18-fix-up-event-handling-with-nested-components)

React/nested-event-handling.md
```diff
@@ -124,6 +128,43 @@ Now let's take a look at an updated version of the diagram that we created in ou
...
+Ultimately, we can refactor the code to look like this:
+
+```javascript
+  // Student.js
+  const Student = (props) => {
+  return (
+    <section>
+      <h3>Student Component</h3>
+      <h4 className={attendanceClass}>Name {props.fullName} </h4>
+      <p>Email: {props.email} </p>
```
droberts-ada: The ordering here feels a little weird, given that the next section talks about displaying attendance as though we haven't thought about it yet.

Also, `attendanceClass` is undefined on line 151.

droberts-ada: I can think of two ways to proceed:
- Here, make it a true refactor (remove the `attendanceClass` and the CSS)
- Here talk about "let's include some code to show whether the student is present", and below change the text to focus specifically on marking a student present.

React/nested-event-handling.md
```diff
@@ -124,6 +128,43 @@ Now let's take a look at an updated version of the diagram that we created in ou
...
+  );}
+```
+
+```css
+   /*Student.css*/
+  .absent {
+    color: red;
+  }
+
+  .present {
```
droberts-ada: Do we have a naming convention for classes used for styling with React? I think the answer is no, but we probably should. Maybe something like, the outermost element of the component gets a class that is the component's name (`.student`), and then all styles reference both the component class and the modifier class (`.student .absent`).

Or just use BEM.

The reason I bring it up is, even though the CSS looks modular it does get applied to the entire site. I don't think we're likely to infect some other component using `.present` or `.absent` in this app, but it might be a risk for some of what the students are doing.

React/nested-event-handling.md
```diff
@@ -134,51 +175,80 @@ There are a few things to consider when making this change to our application. T
 
 If you answered NO to question #3, you're on to something big. The data related to the students is tracked in the `StudentCollection` while the button to mark an individual student "present" should really be on each individual `Student` component. Our challenge then is to use the tools we have been given to _propagate_ the button press event from one component to another.
 
-#### The button
+#### The Button
 Let's start by updating the `render` function of the `Student` component to include the `isPresent` property, and a button.
 
 ```js
-<p>
```
droberts-ada: Would it make sense to change this to not reference the `render` function, since `Student` is no longer a class?

React/nested-event-handling.md
```diff
@@ -134,51 +175,80 @@ There are a few things to consider when making this change to our application. T
...
-  { this.props.isPresent ? 'Present' : 'Absent' } today
-  <button
+// Student.js
+const Student = (props) => {
+
+return (
+  <section>
+    <h3>Student Component</h3>
+    <h4 className={(props.isPresent) ? 'present' : 'absent'}>Name {props.fullName} </h4>
+    <p>Email: {props.email} </p>
```
droberts-ada: Above you use a local variable `attendanceClass` for the `className`.

React/nested-event-handling.md
```diff
@@ -134,51 +175,80 @@ There are a few things to consider when making this change to our application. T
...
-```javascript
+```js
 // Student.js
-
+//...
+return (
 <button
   disabled={ this.props.isPresent }
   onClick={ this.onPresentButtonClick }
-  >Mark Present</button>
```
droberts-ada: Since `onPresentButtonClick` is defined inline, you should omit the `this`.

React/nested-event-handling.md
```diff
@@ -134,51 +175,80 @@ There are a few things to consider when making this change to our application. T
...
+
+<details>
+<summary>Answer</summary>
+`onPresentButtonClick` is closed around `props`, so the `props` variable is still accessible!
+</details>
 
 #### Callback from the Parent
 
-Next, we must consider that the `state` of the students is not stored within this component. In order to update this, we'll need to call some code within the `StudentCollection` component which will change the state.
+Next, we must consider that the `state` of the students is not stored within this component. In order to update this, we'll need to _implement_ some code within the `StudentCollection` component which will change the state, and then _call_ that code inside the `Student`
```
droberts-ada: I like this distinction!

React/nested-event-handling.md
```diff
@@ -190,11 +260,14 @@ markPresent = (studentIndex) => {
 
 **Question:**  What way do we have to pass data from a parent component to a child component.
 
-If you answered `props` you're correct!
+<details>
+<summary>Answer</summary>
```
droberts-ada: Could you make the `**Question:**` line be the summary? You'll have to replace the asterisks with a `<b>`.

React/nested-event-handling.md
```diff
@@ -217,10 +290,14 @@ Now, in the `Student` component, we can call this callback function within our e
 
 ```javascript
 // Student.js
+const Student = (props) => {
+  const onPresentButtonClick = () => {
+    console.log(this);
+    this.props.markPresentCallback(this.props.index);
```
droberts-ada: No `this` in a functional component!

droberts-ada: You'll need to update the snowman diagram below as well.

React/nested-event-handling.md
dHelmgren: Technically we do use BEM but frankly BEM is for people that don't understand how CSS selectors work.

I can go with the `.student .present` style, I follow the logic. It might be running back through the project series but * shrug *.

React/nested-event-handling.md
```diff
@@ -134,51 +165,105 @@ There are a few things to consider when making this change to our application. T
...
 Then we tie the button to the event handler function:
 
-```javascript
+```js
 // Student.js
-
+//...
+return (
 <button
   disabled={ this.props.isPresent }
```
dHelmgren:
```suggestion
  disabled={ props.isPresent }
```

PR #593, opened 2019-07-26
Kw args

jmaddox19:
## Description
A few sentences describing the overall purpose of these textbook changes.

## Questions for the Author
- [ ] Have you updated the weekly learning goals in the pedagogy resource? **N/A**
- [ ] Have you updated the weekly assessment in Schoology? **No, should I?**
- [ ] Have you added links to the source (lucidchart, draw.io) for any diagram? **N/A**
- [ ] Have you updated the READMEs with links to new or moved resources? **N/A**

## Todos
None


01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -157,39 +205,38 @@ coffee_price(:drip, :tall, cold: true)
...
-There are a few things to note here:
-- In the method signature, each keyword argument comes with a _default value_
-  - The syntax is similar to keys and values in a hash
-- When you invoke a method, keyword arguments are optional
-  - If you omit one, the value from the method signature will be used instead
-- When a method is invoked, any keyword arguments are passed by name (a.k.a. _keyword_)
-- Keyword arguments can be passed in any order.
-- Keyword arguments can co-exist with regular arguments (also called _positional arguments_)
-  - Positional arguments must come first, both in the definition and when the method is called
-  - All positional arguments must be supplied for every call
```
kaidamasaki: Are we sure we want to get rid of this list?  It seems like the repetition might be good.

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -157,39 +205,38 @@ coffee_price(:drip, :tall, cold: true)
...
+In each of these iterations, the only things we've changed are the parameters in the method signature, the body of the method remains the same. The end result should resemble [this gist](https://gist.github.com/droberts-ada/07714ad0e2a3bc5a60c50d6a78e02f15).
```
kaidamasaki: We should probably see if we can make this Gist on an Ada account or port it to a current instructor.

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -157,39 +205,38 @@ coffee_price(:drip, :tall, cold: true)
...
 
-The difference between the two is similar to the difference between options (flags) and arguments to a command-line program. Options change _how_ the program behaves, arguments change _what_ it acts on.
+When you make a method, and you write code to call it, ask yourself, "is it hard remember or understand which argument is which?" or conversely "is it more trouble than it's worth to type out the keywords?"
 
 ## Summary
 
+|   | Required | Optional |
+|---|----------|----------|
+| **Positional** | coffee_price(type, size, extra_shots, cold) | coffee_price(type, size, extra_shots = 0, cold = false) |
+| **Keyword** | coffee_price(type:, size:, extra_shots:, cold:) | coffee_price(type, size, extra_shots: 0, cold: false) |
```
kaidamasaki: Nit: Can we line these up with spaces and `-`s so that we can make it look more like an ascii-art table?

01-ruby-fundamentals/keyword-arguments.md
jmaddox19: Ah, it hadn't occurred to me that we'd want repetition within the same lesson/markdown but I'm happy to bring it back.
Much of it would need to be updated (lots of conflating between optional and keyword arguments) but I can certainly do that.

01-ruby-fundamentals/keyword-arguments.md
jmaddox19: Happy to do that. I'm noticing this is just a copy/paste of what's in this file. Do we find it helpful to have the code in it's own file as well to make it easier to find or something?

01-ruby-fundamentals/keyword-arguments.md
jmaddox19: I want to clarify that it looks pretty in an actual markdown editor, so I didn't think of it as worth the effort, but I certainly can

jmaddox19: Screen Shot 2019-07-29 at 11 12 00 AM

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -129,16 +129,64 @@ coffee_price(:latte, :venti, 3, false)
...
+# Price for a tall iced drip
+coffee_price(:drip, :tall, extra_shots: 0, cold: true)
+
+# Price for a grande cappuccino
+coffee_price(:cappuccino, :grande, extra_shots: 0, cold: false)
+```
+
+There are a few things to note here:
+- When a method is invoked, any keyword arguments are passed by name (a.k.a. _keyword_)
+- Keyword arguments can be passed in any order. (ie. `coffee_price(:drip, :tall, cold: true, extra_shots:2`))
```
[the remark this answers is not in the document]
droberts-ada: [This should be "e.g." not "i.e."](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/).

>E.g. stands for exempli gratia and means “for example.” I.e. is the abbreviation for id est and means “in other words.”

I always think of "for <b>eg</b>gsample" (don't ask me why) and "<b>i</b>n <b>e</b>ssence"

01-ruby-fundamentals/keyword-arguments.md
droberts-ada: Not all of these still apply, and some of them have or could be moved elsewhere. I would go through and make sure that none of the points are lost, and that they're all in the right spots.

droberts-ada:
This looks good in general - I really like the new flow of this lesson where you split the two concepts (optional vs keyword) apart, introduce them one by one and then combine them. Just a bit of detail work that needs to be addressed.

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -129,16 +129,64 @@ coffee_price(:latte, :venti, 3, false)
...
+
+### Optional Keyword Arguments
+
+Why not utilize both concepts? We can also make keyword arguments optional by simply adding a default value after the colon.
```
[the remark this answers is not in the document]
jmaddox19: Thanks so much for the feedback and thorough explanation here! I will be more mindful of that going forward.

01-ruby-fundamentals/keyword-arguments.md
jmaddox19: It is all covered elsewhere, a couple weren't covered until the summary so I'll add them after they're introduced as well.

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -129,16 +129,64 @@ coffee_price(:latte, :venti, 3, false)
...
+
+# Price for a venti latte with 3 extra shots
+coffee_price(:latte, :venti, 3)
+
+# Price for a tall iced drip
+coffee_price(:drip, :tall, 0, true)
+
+# Price for a grande cappuccino
+coffee_price(:cappuccino, :grande)
+```
```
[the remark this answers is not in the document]
jmaddox19: Will add an observations paragraph!
In reference to your 2nd paragraph: I considered calling that out but decided against it because I feared that without having introduced keyword arguments yet, students would struggle to follow the problem. Though convenient, it feels to me like more of a by-product of using the keyword and optional concepts together rather than the core reason either was introduced. 
That said, I think you worded it in about as comprehensible of a way as one could and I'm happy to add it there, just thought I'd give my concerns.

01-ruby-fundamentals/keyword-arguments.md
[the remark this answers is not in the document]
jmaddox19: Found a helpful explanation article! https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/

jmaddox19:
Also curious whether I should make changes to the weekly assessment. (I'm not even sure where they are.) Also whether that goes in this PR or a separate one. (I guess if it's in a different repo adding them to this PR isn't an option.)

01-ruby-fundamentals/keyword-arguments.md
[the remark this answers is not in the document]
droberts-ada: I think it's probably fine to leave the second paragraph out - issues 1 and 2 are plenty of motivation for the next technique.

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -129,67 +129,125 @@ coffee_price(:latte, :venti, 3, false)
...
-coffee_price(:drip, :tall, cold: true)
+coffee_price(:drip, :tall, 0, true)
 
 # Price for a grande cappuccino
 coffee_price(:cappuccino, :grande)
 ```
 
-The only things we've changed are the parameters in the method signature, the body of the method remains the same. The end result should resemble [this gist](https://gist.github.com/droberts-ada/07714ad0e2a3bc5a60c50d6a78e02f15).
+Notice that not all the arguments in the method above are optional. Optional Arguments can co-exist with Positional Arguments (and often do).
+Positional arguments must come first, both in the definition and when the method is called.
```
droberts-ada: Have we defined "positional argument" at this point?

01-ruby-fundamentals/keyword-arguments.md
jmaddox19: @kaidamasaki Is there a team tenet around how markdown tables are supposed to look? Should we drive toward forming one?

01-ruby-fundamentals/keyword-arguments.md
jmaddox19: Good point! I'll update for that.

01-ruby-fundamentals/keyword-arguments.md
```diff
@@ -129,67 +129,126 @@ coffee_price(:latte, :venti, 3, false)
...
 
 All of these put the onus on the caller to do things right, which is just asking for trouble. **A well-written method is easy to use right and hard to use wrong** - by that standard, this is not good code.
 
-### Keyword Arguments
+To address these issues, we will introduce two new concepts: optional arguments and keyword arguments. We'll use each by itself first, then show how they can be used together to create clean, readable code.
+
+### Optional Arguments
+
+We're going to tackle #3 on the above list first by introducing ruby's support of **Optional Arguments**.
+Also now that we're introducing a different type of argument, it's important that we recognize that the type of argument we've seen so far is generally referred to as a **Positional Argument**. The name will hopefully make more sense later on in the lesson, but for now you just need to know that's what it's called.
```
droberts-ada: It might make sense to move this definition of positional argument to right after line 139.

To address these issues, we will introduce two new concepts: optional arguments and keyword arguments. We'll use each by itself first, then show how they can be used together to create clean, readable code.

Now that we're introducing different types of arguments, it's important that we recognize that the type of argument we've seen so far is generally referred to as a **Positional Argument**. The name will hopefully make more sense later on in the lesson, but for now you just need to know that's what it's called.

jmaddox19: You're so right! That makes so much more sense! Fixing

PR #602, opened 2019-08-06
Change name length conditional to places visited conditional

jmaddox19:
I've seen name length be used to single out people of color so I thought it might be good to use an example that wasn't explicitly singling out people with long names.

### DIFF
--- a/01-ruby-fundamentals/conditional-control-structures.md
+++ b/01-ruby-fundamentals/conditional-control-structures.md
@@ -10,9 +10,9 @@
 
 ```ruby
 puts "How many cities have you been to?"
-city_count = gets.chomp
+city_count = gets.chomp.to_i
 puts "How many countries have you been to?"
-country_count = gets.chomp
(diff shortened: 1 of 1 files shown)


01-ruby-fundamentals/conditional-control-structures.md
```diff
@@ -9,21 +9,21 @@
 ### Using Compound Conditions
 
 ```ruby
-puts "What is your first name?"
-first_name = gets.chomp
-puts "What is your last name?"
-last_name = gets.chomp
+puts "How many cities have you been to?"
+city_count = gets.chomp
```
kaidamasaki: Should this be `gets.chomp.to_i`?

kaidamasaki: (And also below.)

jmaddox19:
I added to_i and actually ran the code I wrote to make sure it works this time

PR #607, opened 2019-08-09
warm-up review activity for scope

beccaelenzil:
## Description
Adding a review resource for scope to be used at the start of the day the day after the scope lesson is given

## Questions for the Author
- [ NA] Have you updated the weekly learning goals in the pedagogy resource?
- [ NA] Have you updated the weekly assessment in Schoology?
- [ NA] Have you added links to the source (lucidchart, draw.io) for any diagram?
- [NA ] Have you updated the READMEs with links to new or moved resources?

## Todos
- [NA ] Section X still needs some more detailed examples
- [ NA] Still need to add "Additional Resources"

## Feedback Requested

### DIFF
--- a/Review/week-1/scope_review.md
+++ b/Review/week-1/scope_review.md
@@ -18,7 +18,7 @@ Review the code below. What do you notice? What do you wonder?
 ```ruby
 
 # find the equation of a line given two points
-def equation(x1, y1, x2, y2)
+def find_line_equation_(x1, y1, x2, y2)
 
   # -- 1 -- 
   # puts "The slope is #{m}."
@@ -48,15 +48,14 @@ def equation(x1, y1, x2, y2)
     puts "y = #{m}x + #{b}"
   end
 
-  return m,b
+  return {slope: m, y_intercept: b}
 end
 
-m, b = equation(2, 4, 3, 5)
+result = equation(2, 4, 3, 5)
 
 # -- 4 --
-# puts "The slope is #{m}."
-# puts "The y-intercept is #{b}."
-
+# puts "The slope is #{result[:slope]}."
+# puts "The y-intercept is #{result[:y_intercept]}."
 
 ```
 
--- a/Review/week-1/scope_review.md
+++ b/Review/week-1/scope_review.md
@@ -1,0 +1,62 @@
+## Lessons and Exercises
+- [Scope](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/01-ruby-fundamentals/scope.md)
+
+## Warm-up Exercise
+
+Review the code below. What do you notice? What do you wonder?
+
+### Check for understanding
+1. What is the method name?
+2. When do we call the method?
+3. What are the method parameters?
+4. What is the method's return value? What is the data type of this return value?
+5. What happens if we uncomment #1?
+6. What happens if we uncomment #2?
+7. What is the output of #3?
+8. What is the output of #4?
+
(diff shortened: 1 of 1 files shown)


Review/week-1/scope_review.md
```diff
@@ -0,0 +1,62 @@
...
+  elsif b == 0
+    puts "y = #{m}x"
+  else 
+    puts "y = #{m}x + #{b}"
+  end
+
+  return m,b
+end
+
+m, b = equation(2, 4, 3, 5)
```
tildeee: We actually don't teach this cool Ruby thing in our curriculum!

Because we don't teach it, it may be best to refactor this code...

Maybe one thing you could do is have the `equation` method return a hash that has key-value pairs for slope and y-intercept.

So the end of the method would be instead of `return m,b`:
```ruby
return { "m" => m, "b" => b }
```

and we would instead grab the result with
```ruby
result = equation(2, 4, 3, 5)
m = result["m"]
b = result["b"]
```

Or something similar? Anyway, unfortunately most of our students won't know where the `return m, b` or `m, b = equation()` comes from

Review/week-1/scope_review.md
```diff
@@ -0,0 +1,62 @@
...
+4. What is the method's return value? What is the data type of this return value?
+5. What happens if we uncomment #1?
+6. What happens if we uncomment #2?
+7. What is the output of #3?
+8. What is the output of #4?
+
+```ruby
+
+# find the equation of a line given two points
+def equation(x1, y1, x2, y2)
```
tildeee: i think method names are much clearer when they're a _verb_ ish, or action. Could we rename this to something like `find_equation` or `find_line_equation`?

Review/week-1/scope_review.md
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  return m,b
+end
+
+m, b = equation(2, 4, 3, 5)
+
+# -- 4 --
+# puts "The slope is #{m}."
+# puts "The y-intercept is #{b}."
```
tildeee: Why are the contents of this no. 4 commented out? As is, I don't think it should break the code

Review/week-1/scope_review.md
```diff
@@ -0,0 +1,62 @@
...
+```ruby
+
+# find the equation of a line given two points
+def equation(x1, y1, x2, y2)
+
+  # -- 1 -- 
+  # puts "The slope is #{m}."
+  # puts "The y-intercept is #{b}."
+
+  # -- 2 --
```
tildeee: Unless I'm reading this incorrectly, could we leave this code snippet uncommented?

Review/week-1/scope_review.md
beccaelenzil: I'll make this change.

Review/week-1/scope_review.md
beccaelenzil: I'll make this change.

Review/week-1/scope_review.md
beccaelenzil: I was commenting these so if students copied and pasted the code to run it, they didn't see the answers to the questions, but could still experiment with the code.

Review/week-1/scope_review.md
beccaelenzil: I was commenting these so if students copied and pasted the code to run it, they didn't see the answers to the questions, but could still experiment with the code.

PR #630, opened 2019-08-19
Csfun binary numbers

CheezItMan:
## Description

Moved the Binary lesson from just a placeholder linking to slides to a textbook resource.  It's also focusing on how computers work and why they use binary numbers with a small tangent on hex and Unicode. 

## Questions for the Author
- [x] Have you updated the weekly learning goals in the pedagogy resource?

## Todos
- [ ] Video will be recorded soon

## Feedback Requested


04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+
+A _Computer_ is a device for storing and processing data according to a series of commands or instructions provided to it.  It is normally structured with a Central Processing Unit, system memory and an I/O bus, which is often called a _motherboard_ connecting them to a variety of peripheral devices (mice, keyboards, etc).  
+
+![Computer System Example](images/computer-system.png)
+
+| Term 	| Definition 	|
+|---	|---	|
+| CPU 	| The **C**entral **P**rocessing **U**nit of a computer.  This is the main processor which executes program instructions, performs arithmetic and coordinates the other components of a computer. 	|
+| Peripherals 	| External components to a computer which can provide input and/or output.  An example is a monitor or keyboard. 	|
+| MAR 	| The Memory Address Register is where the CPU stores an address of an instructor or data in memory.  When data is transferred to or from RAM this is the address used by the CPU.  |
```
kaidamasaki:
```suggestion
| MAR 	| The Memory Address Register is where the CPU stores an address of an instruction or data in memory.  When data is transferred to or from RAM this is the address used by the CPU.  |
```

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+| MDR | The Memory Data Register is where the CPU stores an information while it is transferred to or from RAM.  |
```
kaidamasaki: Introducing the concept of a page might help clarify this:

```suggestion
| MDR | The Memory Data Register is where the CPU stores a page while it is transferred to or from RAM.  |
```

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+| IR 	| The Instruction Register is where the CPU stores the current command to execute.  |
+| PC 	| The Program Counter is where the CPU stores the address of the next instruction to execute.  |
+| Cache 	| A type of very fast memory, close to or on the CPU chip which is used to store information the CPU will likely need to access soon.  It can also be a term for storing information likely to be used in a location which can be quickly accessed.  This technique of pre-staging information likely to be used in proximate locations is called _caching_. 	|
+| Accumulator 	| The accumulator is used to store an input or output operand during execution of the instructions. Operand is the data to be operated by the computer. 	|
+| Register 	| A type of very fast memory, close to or on the CPU chip which is used to store information the CPU will likely need to access soon.  It can also be a term for storing information likely to be used in a location which can be quickly accessed.  This technique of pre-staging information likely to be used in proximate locations is called _caching_. 	|
```
kaidamasaki: This should probably be moved up above the definitions of the specific registers.

Also, this seems to just be a copy/paste of cache.
```suggestion
| Register 	| A type of very small and fast memory on the CPU chip which is used to store information the CPU is currently using. |
```

04-cs-fundamentals/classroom/Binary.md
kaidamasaki: Oh, nevermind this isn't page related.  The MAR and MDR are definitely not things I remember from school, btw.  (We tended to mention registers more generally but stayed at a page level abstraction when talking about copying memory to/from RAM.)

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+
+### A Note On Subscripts
+
+When we use subscripts like the 2 in 1101<sub>2</sub> to indicate that the number is a binary number.  Likewise we could use 10 as a subscript for a decimal number. 135<sub>10</sub>
+
+## Bits & Bytes
+
+Computers organize binary digits into units called _Bits_.  A _Bit_ is one binary digit, either a 0 or a 1.  Each _memory cell_ is an electronic circuit which stores one bit of data.  However one bit by it's lonesome isn't terribly useful, so the memory cells are grouped into units of 8, called a _Byte_.  In the early days of computers the smallest unit you could work with on a computer was a byte, but in modern computers bytes are too small, and so the smallest individual unit of memory that you can work with is called a _word_.  That means at each memory address is one word of data, which could be multiple bytes in size.  On a 32-bit platform a word is usually 4 bytes in size (32 bits).  The larger the word size, the more possible values it can hold.  The number of values possible to a `k` digit binary number is 2<sup>k</sup>.
```
kaidamasaki: Nobody uses 32 bit computers anymore.  :-P

```suggestion
Computers organize binary digits into units called _Bits_.  A _Bit_ is one binary digit, either a 0 or a 1.  Each _memory cell_ is an electronic circuit which stores one bit of data.  However one bit by it's lonesome isn't terribly useful, so the memory cells are grouped into units of 8, called a _Byte_.  In the early days of computers the smallest unit you could work with on a computer was a byte, but in modern computers bytes are too small, and so the smallest individual unit of memory that you can work with is called a _word_.  That means at each memory address is one word of data, which could be multiple bytes in size.  On a 64-bit platform a word is usually 8 bytes in size (64 bits).  The larger the word size, the more possible values it can hold.  The number of values possible to a `k` digit binary number is 2<sup>k</sup>.
```

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+
+From there bytes are grouped into larger and larger groups.  You may have heard about _Kilobytes_, _Megabytes_, and _Terabytes_ etc.  A KiloByte is 2<sup>10</sup> Bytes or 1024.  A Megabyte is 2<sup>20</sup> Bytes, and so on.
```
kaidamasaki:
```suggestion
From there bytes are grouped into larger and larger groups.  You may have heard about _Kilobytes_, _Megabytes_, and _Gigabytes_ etc.  A KiloByte is 2<sup>10</sup> Bytes or 1024.  A Megabyte is 2<sup>20</sup> Bytes, and so on. 
```

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+
+</details>
+
+## ASCII & Unicode
+
+If everything in a computer is a binary number, how does a computer handle text?  Enter ASCII & Unicode.  [ASCII or the American Standard Code for Information Exchange](https://en.wikipedia.org/wiki/ASCII) was published in 1963 and is a table with text characters represented by a number.  In the early days of comptuters when the sizes of variables were small, there were a very limited number of possible characters which could be represented resulting in this table.
+
+![ASCII Table](images/ascii-table.png)
+
+Originally these were the standard characters a computer could represent.  The table was limited in size because characters were assumed to be 1 byte in length.  Today ASCII has largely been replaced by [Unicode](https://home.unicode.org/) a which much larger set of characters and multiple methods of encoding which allows a variety of languages and even emojis.  ASCII however lives on as a subset of Unicode & in [wonderful ASCII Art](https://www.asciiart.eu/).
```
kaidamasaki: emojis -> emoji (it's already plural)

Also, I've added a line about different encodings.

```suggestion
Originally these were the standard characters a computer could represent.  The table was limited in size because characters were assumed to be 1 byte in length.  Because ASCII only really supported English text different countries came up with a variety of different encodings.

Today ASCII and other encodings have largely been replaced by [Unicode](https://home.unicode.org/)(**uni**versal **code**) a which much larger set of characters and multiple methods of encoding which allows a variety of languages and even emoji.  ASCII however lives on as a subset of Unicode & in [wonderful ASCII Art](https://www.asciiart.eu/).
```

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+
+**Sorting** Usually when you compare strings in a programming language you end up comparing the numeric codes of each letter.  So "A" will be consider less than "a" (65 compared to 97).
```
kaidamasaki: Unicode and foreign languages makes this more complicated.  (Sorting things like å and whatnot.)

```suggestion
**Sorting** Usually when you compare English strings in a programming language you end up comparing the numeric codes of each letter.  So "A" will be consider less than "a" (65 compared to 97).
```

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+
+## Hexidecimal
+
+Because binary numbers can be a hassle to write because even a small number can require many digits, sometimes programmers will use base 16, or hexidecimal numbers.  Because 16 is 2<sup>4</sup>, each Hexidecimal digit can be represented by 4 binary digits.
+
+**Hex**|**Binary**
```
kaidamasaki: Adding a column for decimal could be handy here.

04-cs-fundamentals/classroom/Binary.md
```diff
@@ -1,22 +1,277 @@
...
+```c
+#include <stdio.h>
+int main()
+{
+   printf("Hello, World!");
+   return 0;
+}
+```
+
+Other languages like Ruby or Python are [interpreted languages](https://en.wikipedia.org/wiki/Interpreted_language) these languages are run by a special program called an _[interpreter](https://en.wikipedia.org/wiki/Interpreter_(computing))_ which reads the program and tells the computer what to do.  Because compiled languages are executed in machine-code form, they are often faster than interpreted languages, but an interpreted language does not need to be compiled for a specific type of computer.  An interpreted language can run on any computer with a valid interpreter.
```
kaidamasaki: It might be worth mentioning that some languages can be interpreted _or_ compiled, since the split between interpreted and compiled languages can be a little fuzzy.  

It also could be worth mentioning a language like Java which is compiled into byte-code which is then interpreted.

PR #661, opened 2019-09-16
Rails Smoothing Week 1 Day 1&2

dHelmgren:
## Description
I'm smoothing out Rails week 1! I'm making 5 PR's, one for each day, so that it isn't super overwhelming to review! These changes are meant to smooth transitions between lessons by improving consistency and making sure all errors are addressed!

I messed up so monday and tuesday are in this one WHOOPS

### DIFF
--- a/08-rails/show-action-intro.md
+++ b/08-rails/show-action-intro.md
@@ -51,13 +51,19 @@ We'll need to add a new method, `show`. Inside that method we'll be able to acce
 # app/controllers/books_controller.rb
 def show
   book_id = params[:id].to_i
-  @book = BOOKS[book_id]
+  if book_id < BOOKS.length
+    @book = BOOKS[book_id]
+  else
+    head :not_found
+  end
 end
 ```
 
 Here we read the book ID from the params and store it in a variable `book_id`, then use that as an index into our list of books. As before, we'll use an instance variable to communicate with the view, though this time we've only got one `@book`, not multiple `@books`.
 
 **Note:** In this implementation, we need to explicitly call `.to_i` on `params[:id]` because we need to use `book_id` as an index on the `BOOKS` array. We won't necessarily need to do this `.to_i` call in the future-- stay tuned!
+
+Finally, there is that little bit at the end with `head :not_found`. We want to make sure that we are doing something meaningful when there isn't something to show, so this line sends back an error code 404, or 'not found' in laymen speak. We'll talk about more robust ways to deal with this in the future, but for now it's a good enough placeholder.
 
 ### Exercise: View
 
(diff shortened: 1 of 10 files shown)


dHelmgren:
ahh shoot

08-rails/using-active-record-in-code.md
```diff
@@ -0,0 +1,87 @@
...
+  end
+```
+</details>
+
+## Active Record Objects in ERB
+
+While this will help us use Active Record in the controller, we still need to make some adjustments. Since we're no longer using a hash, we need to remove the code that looks like this:
+
+```ruby
+<%= @book[:author]%>
```
kaidamasaki: I think there should be a space before the `%>`

```suggestion
<%= @book[:author] %>
```

08-rails/using-active-record-in-code.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+Luckily, since we are using Active Record, this is as simple as treating the data as what it is: an object! The code above becomes:
+
+```ruby
+<%= @book.author%>
```
kaidamasaki:
```suggestion
<%= @book.author %>
```

dHelmgren:
[Rails Smoothing: Week 1 Monday & tuesday](https://trello.com/c/R8Bjja2P/1755-rails-smoothing-week-1-monday-tuesday)

08-rails/show-action-intro.md
```diff
@@ -50,15 +50,21 @@ We'll need to add a new method, `show`. Inside that method we'll be able to acce
...
+    return
+  end
 end
 ```
 
 Here we read the book ID from the params and store it in a variable `book_id`, then use that as an index into our list of books. As before, we'll use an instance variable to communicate with the view, though this time we've only got one `@book`, not multiple `@books`.
 
 **Note:** In this implementation, we need to explicitly call `.to_i` on `params[:id]` because we need to use `book_id` as an index on the `BOOKS` array. We won't necessarily need to do this `.to_i` call in the future-- stay tuned!
 
+Finally, there is that little bit at the end with `head :not_found`. We want to make sure that we are doing something meaningful when there isn't something to show, so this line sends back an error code 404, or 'not found' in laymen speak. We'll talk about more robust ways to deal with this in the future, but for now it's a good enough placeholder.
```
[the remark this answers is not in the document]
dHelmgren:
```suggestion
Finally, there is that little bit at the end with `head :not_found`. We want to make sure that we are doing something meaningful when there isn't something to show, so this line sends back an error code 404, or 'not found' in lay-speak. We'll talk about more robust ways to deal with this in the future, but for now it's a good enough placeholder.
```

PR #663, opened 2019-09-18
Dsh/rails smoothing day5

dHelmgren:
## Description
Last one done! I've got a change that goes through some of the older material here, so we need to merge it last.


dHelmgren:
[Rails Smoothing: Week 1 Friday](https://trello.com/c/th5XKATW/1761-rails-smoothing-week-1-friday)

08-rails/routes-and-resources.md
```diff
@@ -2,53 +2,62 @@
...
 
-## Named Routes for REST
-Since the standard RESTful resource only requires two paths, all five actions can be represented by just the two named routes.
+```ruby
+get "/books/new", to: "books#new"
+get "/books/:id", to: "books#show"
+```
 
-The conventional route names for this RESTful resource would be `books_path` and `book_path`. We use the plural version when referring to a _collection_ of Book objects (`index` action), and the singular version when referring to an _instance_ or _specific_ Book (`show`, `edit`, etc).
+That general rule is that routes go from most at the top, to least specific at the bottom for any given resource.
```
kaidamasaki: This reads a little confusingly to me.

dHelmgren: And no wonder, I forgot a.
```suggestion
That general rule is that routes go from most specific at the top, to least specific at the bottom for any given resource.
```

PR #681, opened 2019-10-24
forEach Loop and for-in Loop

CheezItMan:
## Description

I added a bit of examples about the for-in loop and how you can't break out of a `forEach` loop.

## Questions for the Author
- NA

## Todos

- NA

## Feedback Requested

### DIFF
--- a/10-JavaScript/anonymous-and-for-each.md
+++ b/10-JavaScript/anonymous-and-for-each.md
@@ -112,6 +112,58 @@ const average = sum / cashTips.length;
 console.log(`The average cash tip is ${average}`);
 ```
 
+## Danger Will Robinson Danger
+
+There is one thing to note about the `forEach` loop.  Consider this linear search:
+
+```javascript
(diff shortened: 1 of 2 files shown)


10-JavaScript/anonymous-and-for-each.md
```diff
@@ -112,6 +112,58 @@ const average = sum / cashTips.length;
...
+```
+Alice is not Cooper
+Found it!
+Maria is not Cooper
+Alyokhina is not Cooper
+Alicia is not Cooper
+Keys is not Cooper
+```
+
+You **cannot** break out of a `forEach` loop.  It will iterate throughout the collection.  This can be good, if this is the behavior you want, but if you want to short-circuit the loop, [another method might be preferable.](http://frontendcollisionblog.com/javascript/2015/08/15/3-reasons-you-should-not-be-using-foreach.html) Just remember if you are in a `forEach` loop, you are in for the full ride.
```
kaidamasaki: We probably don't want to mention breaking out of a `.forEach` with an exception, but it's definitely possible.

PR #684, opened 2019-11-05
Remove the CSRF file update

CheezItMan:
## Description

I just discovered that we'd had a line in our Post request lesson that no longer applies to current versions of Rails 5 or 6.  So I'm taking it out and making a note that you'd have to do something different if you had an App which served both HTML and JSON.


09-intermediate-rails/api-creating-items.md
```diff
@@ -62,14 +62,7 @@ As before, the big difference from a web app is in what we send back. While in a
...
 
-Since we expect requests to come from other sites/applications in an API context, we will add a Rails helper method to our controller to allow these `POST` requests to go through:
-```ruby
-# pets_controller.rb
-class PetsController < ApplicationController
-  protect_from_forgery with: :null_session
-
-  ...
-```
+We don't this in an API application, although we might need to modify our code if we wrote an application which could serve both API and browser requests.
```
beccaelenzil: Should this say "We don't do this..." ?

PR #701, opened 2020-01-09
Update template for Rails 6

kaidamasaki:


kaidamasaki:
Haven't tested this yet, just wanted to get the code up here.

08-rails/reference/.rails-template.rb
```diff
@@ -114,6 +137,28 @@
     RUBY
   end
 
+   gsub_file "test/test_helper.rb", 'parallelize(workers: :number_of_processors)' do
+    "# parallelize(workers: :number_of_processors) # causes out of order output."
```
kaidamasaki: If I can remember how I implemented a parallel test runner with ordered output for RSpec at Socrata I might contribute a bug fix because running in parallel is nice.  (You had to manually opt in there though, so the technique might not work here.)

08-rails/reference/.rails-template.rb
```diff
@@ -10,64 +12,87 @@
...
+  gem_group :development do
+    # Improve the error message you get in the browser
+    gem "better_errors"
+
+    # Nice interactive terminal when an exception happens
+    gem "binding_of_caller"
+
+    # Automatically run our tests
+    gem "guard"
+    gem "guard-minitest"
```
CheezItMan, line 47: Would guard be ok in an API?

kaidamasaki: Oh, yeah.  Moved that by accident.  Good catch!

kaidamasaki:
@CheezItMan I think when I was writing this I wasn't sure if we were doing yarn or npm and the tutorials online used yarn.

I'll update to use npm for this.

kaidamasaki:
Ha, just kidding.  Rails uses `yarn` internally to install webpack.

PR #704, opened 2020-02-04
adds details on what a block param is

tildeee:


01-ruby-fundamentals/nesting-arrays-hashes.md
```diff
@@ -84,16 +84,63 @@ The allergy information for each meal is an array inside a hash inside an array!
...
+    - Again, to access the value of each element in the array, we reference it by the block parameter, or `item` in this case
+    - Is `item` an array? Then you can use all of the good array syntax to access/read it!
+    - Is `item` a hash? Then you can use all of the good hash syntax to access/read it!
+    - If you're unsure of what data type `item` is, do some debugging work to find out
+
+</details>
+
+### Iteration and Nesting
+
+Within our _iteration block_, we should write all of the lines of code that need to execute per element. If we need to print something per `item`, we should write `puts item` here. If we need to create variables, store data, modify data, and/or create data per element, we should do that here.
```
dHelmgren:
```suggestion
Within our _iteration block_, we should write all of the lines of code that need to be executed for each element. If we need to print something per `item`, we should write `puts item` here. If we need to create variables, store data, modify data, and/or create data per element, we should do that here.
```

01-ruby-fundamentals/nesting-arrays-hashes.md
```diff
@@ -84,16 +84,63 @@ The allergy information for each meal is an array inside a hash inside an array!
...
+
+Similarly, if you need to access a deeper piece of data based on each `item`, then you should use `item` appropriately with the techniques we practiced today.
+
+Even if `item` is one element within the array of `menu`, `item` may represented an array or hash itself.
+
+#### Iteration Tips
+
+- Proper/improper indentation in the code can make or break you understanding of your code
```
dHelmgren:
```suggestion
- Proper/improper indentation in the code can make or break the readability of your code
```

01-ruby-fundamentals/nesting-arrays-hashes.md
```diff
@@ -84,16 +84,63 @@ The allergy information for each meal is an array inside a hash inside an array!
...
+- We do not `return` anything within iteration using `each` loops. In `each` loops, we are simply writing code to execute. We use `return` in order to return a value from a method, or in the future we'll use `return` for very specific pieces of syntax
```
dHelmgren: Might be worth it to talk about `break`?

01-ruby-fundamentals/nesting-arrays-hashes.md
```diff
@@ -84,16 +84,63 @@ The allergy information for each meal is an array inside a hash inside an array!
...
+- We can also use `each` in hashes. However, the syntax will need to change, as the block parameter in an `each` loop on a hash will not represent every _element_, but every _key-value pair._ You may need to look up online what this syntax is in Ruby.
```
dHelmgren:
```suggestion
- We can also use `each` to iterate over hashes. However, the syntax will need to change, as each item in a hash is not a single _element_, but instead a _key-value pair_. You may need to look up online what this syntax is in Ruby.
```

PR #708, opened 2020-02-12
removes CSV write as mandatory, adds structure to building CSV.read in a meaningful way

tildeee:
Big things to note:

1. REMOVES PRETTY MUCH ALL OF THE WEIRD FORMS OF SYNTAX and focuses on giving ONE recommendation: using `read`, headers, and an each loop
1. More focus on concept, then syntax
1. Makes CSV write material optional. Only do this if there's time and interest.

### DIFF
--- a/02-intermediate-ruby/csv-as-a-database.md
+++ b/02-intermediate-ruby/csv-as-a-database.md
@@ -1,11 +1,54 @@
 # CSV as a Database (DB)
-## Learning Goals
-- Learn a new approach to organizing data
-- Explore _comma separated values (csv)_ as a storage pattern for data.
-- Discuss how to use a class to model a collection of data
 
-### What is a Database?
-A _database_ is a system of storing and organizing information on a computer. In this case a _CSV_ will act as our database. In most cases a database can be thought of as a table or group of tables:
+## Lesson Pre-Requisites
+
+- This lesson will use iteration and nested arrays and hashes
+- This lesson will reference `CSV` as a class, although classes are not necessary
+- This lesson uses `awesome_print` to clarify the printing of arrays and hashes. Make sure it's installed with `$ gem install awesome_print`
+
+## Learning Goals
+- Learn the concept of separating our data source into a different system (such as a database)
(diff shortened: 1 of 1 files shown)


02-intermediate-ruby/csv-as-a-database.md
```diff
@@ -1,11 +1,54 @@
...
+
+## What is a Database?
+
+A _database_ is a system of storing and organizing information on a computer.
+
+Our goal is to learn a pattern for organizing our code in the following ways:
+
+1. **We will use a CSV file as our database,** or system to store and organize our data
+1. **We will write Ruby code to read the data from the CSV/database**
+1. We will continue to use Ruby code and Ruby methods to analyze, iterate over, and process our data from our database
```
jmaddox19:
```suggestion
1. We will continue to use Ruby code and Ruby methods to analyze, iterate over, and process our data from the database
```

02-intermediate-ruby/csv-as-a-database.md
```diff
@@ -1,11 +1,54 @@
...
+
+## CSV Files Are One Way to Store and Organize Data
+
+In this case a _CSV_ will act as our database.
```
jmaddox19:
```suggestion
In this case a _CSV_ file will act as our database.
```

02-intermediate-ruby/csv-as-a-database.md
```diff
@@ -14,136 +57,292 @@ A _database_ is a system of storing and organizing information on a computer. In
...
+
+## Summary
+
+We should observe a pattern that our **source of data, or database** is a system of organizing data that can be separated from our Ruby code.
+
+We can use Ruby code and `CSV.read` to read the CSV file.
+
+We can use a combination of `CSV.read`, the `headers: true` option, and `.each` in order to iterate through this data.
+
+## Bonus Syntax: `CSV.read` and Forcing `CSV::Table` to be an Array
```
jmaddox19: Will there still be a project and/or activity that requires them to do this? 
If so, I'd prefer we not mark it as bonus to make it clear Instructors should call it out.

PR #710, opened 2020-02-13
first example: changed 8 hours to 11 hours

beccaelenzil:
I changed the first test input from 8 hours to 11 hours, since we are initially not dealing with the issue of needing to prepend 0s. It seems make sense to have a simpler example and then note how this would change for single digit values.


beccaelenzil:
[small change to testing lesson (first example)](https://trello.com/c/LRikSpQU/1961-small-change-to-testing-lesson-first-example)

PR #713, opened 2020-02-18
Dsh/ref val comic clone

dHelmgren:
## Description
I added a new comic type walkthrough of that big example, and then added a bit about `clone` at the end!


dHelmgren:
[Week 4: Add a diagram for the big example in references and values](https://trello.com/c/8FcvBTCf/1712-week-4-add-a-diagram-for-the-big-example-in-references-and-values)

dHelmgren:
@tildeee I think it still will, because the pictures will go by real fast. In fact, I've saved us about 4000 words! ;)

PR #717, opened 2020-02-21
Introduce calling parent constructor separate from adding child properties

jmaddox19:


02-intermediate-ruby/object-inheritance.md
```diff
@@ -57,9 +57,9 @@ This is known as _implicit inheritance_ because when we put functionality in the
 
 ## Super & Initialize
 
-We have seen how inheritance allows a subclass to "inherit" the methods, in this case `mailing_address`, from it's superclass.  We can also create an `initialize` method in the subclass and use it to give our subclass additional attributes.
+Notice the Apartment class is inheriting the mailing_address method from Property but its also inheriting the constructor and all of the reader methods.
```
tildeee:
```suggestion
Notice the Apartment class is inheriting the `mailing_address` method from Property, and it's also inheriting the constructor and all of the reader methods.
```

02-intermediate-ruby/object-inheritance.md
```diff
@@ -79,23 +79,59 @@ class Property
...
+  def initialize(id, street, unit, city, state, zip)
+    super(id, street, city, state, zip)
+    @unit = unit
+  end
+end
+```
+
+Note: Now that we are adding a new attribute to the `Apartment` class, we are required to call super if we want the parent constructor to execute (which is all the time).
+
+`super` **must** be the first line in a subclass' `initialize` method.  If `super` is called without an argument, it will use the parameters from the subclass' `initialize` method.
```
tildeee: :+1: ✨

PR #718, opened 2020-02-21
Add activities throughout lesson with Shape example

jmaddox19:


02-intermediate-ruby/object-inheritance.md
```diff
@@ -53,6 +53,8 @@ fun_place.mailing_address #=> "111 Boardwalk"
                           #   "New York, NY 12070"
 ```
 
+> **Exercise**: Create a `Shape` class with a `color` attribute and a `describe` method. Create a `Circle` class as a child.
```
tildeee: This exercise is immediately underneath "Inheriting Methods," and it just has them creating two classes. Is there a stronger way we can connect the two concepts? Is there a way they can verify that they're on the right track, with the right syntax?

02-intermediate-ruby/object-inheritance.md
```diff
@@ -130,6 +132,8 @@ Note: Now that we are adding a new attribute to the `Apartment` class, we are re
 
 `super` **must** be the first line in a subclass' `initialize` method.  If `super` is called without an argument, it will use the parameters from the subclass' `initialize` method.
 
+> **Exercise**: Update the `Circle` class to have a `radius` attribute.
```
tildeee: Similar to the cmoment above; this is right underneath talking about `super`. Is there a more specific direction that will give students immediate feedback on if they're doing the right thing, whether it's code to run, or sample code to check?

02-intermediate-ruby/object-inheritance.md
```diff
@@ -185,7 +189,35 @@ In this case, we utilize the behavior that the _base class_ (`Property`) provide
...
+  end
+end
+
+fun_place = House.new(56789, '100 Park Place', 'New York', 'NY', 12077, 'Shake')
+fun_place.mailing_address #=> "100 Park Place"
+                  # "New York, NY 12077"
+fun_place.roof_type #=> "Shake"
+```
+
+> **Exercise**: Add `Rectangle` as a child of shape.
```
tildeee:
```suggestion
> **Exercise**: Add `Rectangle` as a child of `Shape`.
```

Conforms the sentence to consistently use all class names

02-intermediate-ruby/object-inheritance.md
```diff
@@ -185,7 +189,35 @@ In this case, we utilize the behavior that the _base class_ (`Property`) provide
 
 ![super in method calls](images/inheritance-super-method.png)
 
-**Exercise**  With your seatmates create a `Condo` class.  In addition to the same properties as `Apartment`, a Condo should also have `price` and `square_feet` properties and a `price_per_square_foot` method.
+> **Exercise**: Override the `describe` method in `Circle` to describe radius as well as `color`.
```
tildeee: Should "radius" also be `radius` in order to be consistent with `color`? I can't really tell your intention from this sentence.

Also, I don't think there was ever a concrete description of what the `describe` method should do in the first place, so I think there might be some students that had no idea that the `Shape`'s `describe` method even mentioned `color`.

02-intermediate-ruby/object-inheritance.md
jmaddox19: Yes, great point. I'll add instructions for that. Thanks!

02-intermediate-ruby/object-inheritance.md
jmaddox19: Yes, great point. I'll add instructions for that. Thanks!

jmaddox19:
@tildeee I agree. Changing the lesson from 90 minutes to 2 hours was on my list of suggestion in the week in review document. 

@beccaelenzil I agree that it would be nice to call back to an example from a previous lesson. I unfortunately don't have the space to update that right now. I wish I'd thought of that when I was writing this lesson.

I had students tell me they learned a lot from this lesson at the end of the week, so I think that means the new example wasn't too much of a barrier for them. But still definitely an area for improvement.

PR #729, opened 2020-03-24
Be/ruby to js solution

beccaelenzil:
## Description
Created a solutions directory. A markdown solution to ruby to js, and a javascript file with the solutions.

## Feedback Requested
I'm wondering if we should use the ruby-to-js-worksheet-solution.md as the main assignment file since the solutions are in details tags. Do we need a separate solution file? or should I remove the solutions from details tags to just have this file be the javascript solutions.

### DIFF
--- a/10-JavaScript/solutions/ruby-to-js-worksheet-solution.md
+++ b/10-JavaScript/solutions/ruby-to-js-worksheet-solution.md
@@ -1,0 +1,97 @@
+# Ruby to JavaScript Worksheet
+Read the code in each section, then write the equivalent JavaScript code for the Ruby that is given. Start by writing it out in a text editor or on a piece of paper. Then try running it and tweak your code until it successfully runs with expected output.
+
+Each problem stands alone. Variables from previous problems do not exist.
+
+## Problem Set
+1. &nbsp;
+    ```ruby
+    person_age = 55
+    ada_age = 2
+
+    if person_age < ada_age
+       print "This person is younger"
+    elsif ada_age < person_age
+       print "Ada is younger"
+    else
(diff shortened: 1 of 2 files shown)


beccaelenzil:
[ruby-to-js solution](https://trello.com/c/XOFiKL0A/1992-ruby-to-js-solution)

10-JavaScript/solutions/ruby-to-js-worksheet-solution.md
```diff
@@ -0,0 +1,246 @@
...
+    <details>
+    <summary>
+    Javascript
+    </summary>
+
+    ```javascript
+    let total = 0
+    const values = [4, 6, 2, 8, 11]
+
+    for(const i in values){
```
jmaddox19: I like that you had a separate foreach example above. Would it be worth adding for these other loop examples as well?

10-JavaScript/solutions/ruby-to-js-worksheet-solution.md
```diff
@@ -0,0 +1,246 @@
...
+      if (x > y){
+        console.log("x is bigger")
+      }
+      else{
+        console.log("x = y")
+      }
+    }
+    else{
+      console.log("y is bigger")
+    }
```
CheezItMan: Minor style suggestion for things like this.  Putting a space before the { is most common.

```suggestion
    if (x > y || x == y) {
      if (x > y) {
        console.log("x is bigger")
      }
      else {
        console.log("x = y")
      }
    }
    else {
      console.log("y is bigger")
    }
```

PR #745, opened 2020-04-07
first pass at questions

dHelmgren:
Any thoughts?


dHelmgren:
I'm gonna leave it as is, partially because we did a demo of the json-server tool today.

PR #752, opened 2020-04-13
adds more details to the existing prompts

tildeee:
Update for C13: there isn't enough time to move this off of mediaranker, but I thought there could be more detail

### DIFF
--- a/React/exercises/identifying-components.md
+++ b/React/exercises/identifying-components.md
@@ -28,7 +28,7 @@ Answer the following questions:
 
 ## Part Two: Make a List of Possible Components with CraigsList
 
-Break into pairs and take a look at [CraigsList](https://seattle.craigslist.org/).
+Take a look at [CraigsList](https://seattle.craigslist.org/).
 
 Identify the visual "components" the website is comprised of, and create a list of components that a React CraigsList would need, just like above.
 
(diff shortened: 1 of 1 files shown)


React/exercises/identifying-components.md
```diff
@@ -1,14 +1,42 @@
...
+- For each component, what is its name, what does it represent?
+- For each component, what website elements are inside of it?
+- For each component, is it a sibling to another component?
+- For each component, are there another components nested inside of it?
+- For each component, does it get used more than once in this page? If yes, what are the differences in UI between each time it's used?
+- Do you agree with this list of components, what they represent, and how they're named? Do you have any disagreements, or other ideas?
+
+## Part Two: Make a List of Possible Components with CraigsList
+
+Break into pairs and take a look at [CraigsList](https://seattle.craigslist.org/).
```
jmaddox19, line 31: Are we planning to facilitate the breaking into pairs? If not, but we do want them to talk with someone about it somehow, I think it'd be helpful to give more direction for C13. 
But also maybe you were planning to leave this as if for future cohorts and communicate about this to students separately. Totally open to that :)

tildeee: My intention was to leave the wording vague in order for this to be useful for C14, but I'll just go ahead and update this now to just "Take a look" without the pairs. Thank you :+1: :smile:
6

PR #757, opened 2020-04-22
index.html.erb file creation

beccaelenzil:
the textbook curriculum said that `index.html.erb` was there, but since we manually created the `index` action, we also needed to manually create this view file.


beccaelenzil:
[small rails textbook curriculum edits](https://trello.com/c/IYk6AfYa/2110-small-rails-textbook-curriculum-edits)

PR #758, opened 2020-04-22
rails db:migrate

beccaelenzil:
If we don't run `rails db:migrate` we get this error. Do you think this is the point where students should do this?

/Users/becca/Documents/GitHub/AdaGold/ada-books/db/schema.rb doesn't exist yet. Run `rails db:migrate` to create it, then try again. If you do not intend to use a database, you should instead alter /Users/becca/Documents/GitHub/AdaGold/ada-books/config/application.rb to limit the frameworks that will be loaded.

### DIFF
--- a/08-rails/live-code-intro.md
+++ b/08-rails/live-code-intro.md
@@ -70,6 +70,7 @@ The last step before we're ready to go is to tell the postgres program running o
 
 ```
 $ rails db:create
+$ rails db:migrate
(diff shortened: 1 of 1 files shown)


beccaelenzil:
[small rails textbook curriculum edits](https://trello.com/c/IYk6AfYa/2110-small-rails-textbook-curriculum-edits)

PR #762, opened 2020-04-24
updated show test

beccaelenzil:
Included updated code for show test so that the test doesn't fail now that the controller/views point to the database.


beccaelenzil:
[updated active record in code to include show test change](https://trello.com/c/wA4uJPuX/2111-updated-active-record-in-code-to-include-show-test-change)

PR #763, opened 2020-04-24
edit and update solution

beccaelenzil:
I added the edit and update solution to a details tag so students have this for reference. I couldn't get the formatting quite right. If anyone's able to fix that formatting and merge it in, that'd be awesome


beccaelenzil:
[updated rails form lesson to include solution for edit/update](https://trello.com/c/cDpcBtzH/2118-updated-rails-form-lesson-to-include-solution-for-edit-update)

PR #773, opened 2020-05-04
begins roundtable discussion: strong params, branches in git

tildeee:


roundtables/rails-week2/day-2-strong-params-branches-in-git.md
```diff
@@ -0,0 +1,67 @@
...
+For this discussion, think of your instructor as a mediator and guide, rather than as a lecturer. You are encouraged to do the following:
+
+* Whenever possible, you should use voice.
+* You should pose questions to the group, not the instructor.
+* Similarly, you are encouraged to answer questions, with the instructor available to correct misconceptions and provide answers when no one else has an answer.
+* Say something, even if you think you might be wrong! (Nobody throws a frisbee well on their first try)
+* When discussing code, you are encouraged to use VS Code Liveshare, and or screen sharing through zoom.
+* Take notes!
+
+## Materials
```
beccaelenzil, line 14:
```suggestion
* [Testing Database Changes in Controllers](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/08-rails/testing-db-in-controllers.md)
* [Routes with Resources](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/08-rails/routes-and-resources.md)
* [Partial Views](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/08-rails/partial-views.md)

tildeee:
@beccaelenzil Good thoughts!

On a personal level, I think that reviewing testing DB changes is not as important as the time for the other subjects, and wasn't on my list of things that needed Roundtable topics. If it's about thoroughness of topics, I think my feeling is "no" on bringing that topic in, but if you have an opinion that the students will need it more than reviewing other material, then I want to consider that.

Resources with Routes was marked as topic I would want reviewed, but compared to the time spent reviewing strong params and branching, I thought it was the least important since it has the least negative consequences if done "incorrectly." Let me know if you have pushback though.

Not sure what to think of the git thing-- I really want them to review git branching individually before they do it on RSR, which is why it's pushed up in the schedule-- I'm hoping to keep it there, unless you think that it will be harmful to the students

jmaddox19:
I only reviewed tomorrow's roundtable. (Wednesday's is now in this PR as well.) 
It looks good to me! 
I like Becca's idea of having people show and tell how they incorporated the lessons into their project but I also like the questions here and that's already a lot of material to cover so I'm in favor of keeping it as is.

PR #788, opened 2020-05-15
require_login filter except current_user action, not only on create

tildeee:
The lesson material didn't have the code that is currently in the ada-books implementation; here's a quick attempt at trying to make them the same and have explanation around it. @jmaddox19 not sure what the content of the video is, so please check to see that this aligns with your video and change it if not!


jmaddox19:
Thanks Dee! This looks great and totally matches the video!

PR #792, opened 2020-05-21
conforms lesson to new pet fields, adjusts some syntax

tildeee:
I changed the pets api so it would have fields consistent with the `ada-pets-node` project, so I've updated the lesson notes to reflect that.

There are also some tiny syntax adjustments according to my preferences, feel free to reject that/change it!

The new Pets API repo: https://github.com/AdaGold/ada-pets-rails-api


09-intermediate-rails/api-creating-items.md
```diff
@@ -221,6 +226,37 @@ describe "index" do
...
+        post pets_path, params: pet_data
+
+      # Assert
+      }.wont_change "Pet.count"
+
+      body = check_response(expected_type: Hash, expected_status: :bad_request)
+      expect(body["errors"].keys).must_include "age"
+    end
+
+  end
```
jmaddox19, line 259: line 192 is written to imply the code snippet is only showing the `index` tests. 
Can we either change line 192 or remove the `show` tests from this snippet?
I think I have a slight preference to keep it to just the `index` tests because I think less code will help reduce distraction and make the message more clear.

